PR #4, opened 2026-02-13
Simple CI

bkontur:


bkontur:
waiting fo SRE guys to do some manual settings to allow runners and stuff

Issue #28, opened 2026-03-03
Rewrite actual zombienet setup to use zombienet-sdk


bkontur:
Also do similar changes for Bulletin:

Some examples:
https://github.com/paritytech/polkadot-bulletin-chain/pull/241
https://github.com/paritytech/polkadot-bulletin-chain/pull/244
https://github.com/polkadot-fellows/runtimes/blob/main/integration-tests/zombienet/src/lib.rs

RafalMirowski1:
@bkontur should we just swap the binary to use zombie-cli from zombienet-sdk (as suggested here: https://github.com/paritytech/web3-storage/pull/29#issuecomment-4003822089) or do we want to spawn zombienet network in rust code?

RafalMirowski1:
When trying to run our zombienet.toml with new binary, I ran into some problems with relative paths in zombienet: https://github.com/paritytech/zombienet-sdk/issues/520

PR #29, opened 2026-03-04
Migrate to zombienet sdk tests

RafalMirowski1:
implements https://github.com/paritytech/web3-storage/issues/28

  - Replace Zombienet (zombienet spawn zombienet.toml) with zombienet-sdk, remove zombienet.toml
  - Add zombienet-sdk-tests crate with Layer 0 (storage), Layer 1 (filesystem, S3) integration tests and a smoke binary for local dev used in just commands
  - Add CI workflow (.github/zombie-tests.yml) with matrix for zombienet-tests

Still missing: reflect changes in docs


justfile
```diff
@@ -139,20 +126,18 @@ generate-chain-spec: build-runtime
...
-    @echo "Setup complete! Run 'just start-chain' and 'just start-provider' to start the local network."
+    @echo "Setup complete! Run 'just start-chain' to start the local network."
 
 # ============================================================
 # File System (Layer 1) Commands
 # ============================================================
 
-# Run the file system basic usage example
+# Run the file system basic usage example (assumes infrastructure is running)
 fs-example:
```
RafalMirowski1, line 136: not sure if i should keep examples as this can be run end-to-end with fs-demo

bkontur:
@RafalMirowski1 @karolk91 @pepoviola dq, why is zombienet-sdk tests better for example than ‎zombienet.toml + PAPI examples? Can we run PET or custom PAPI tests with zombienet-sdk network setup?
image

Can we use zombienet-sdk just to run infra part as an alternative to ‎zombienet.toml? Do we really need to rewrite the PAPI tests to the rust? Why? When we have PET tests, I am sure that we want to rewrite them to the zombienet-sdk.

My understanding of #28 "Rewrite actual zombienet setup to use zombienet-sdk", was really just to rewrite zombient.toml to zombienet-sdk, but just the network setup and not all the tests. Am I missing anything here?

bkontur (changes requested):
https://github.com/paritytech/web3-storage/pull/29#issuecomment-4000608359

RafalMirowski1:
> @RafalMirowski1 @karolk91 @pepoviola dq, why is zombienet-sdk tests better for example than ‎zombienet.toml + PAPI examples? Can we run PET or custom PAPI tests with zombienet-sdk network setup? image
> 
> Can we use zombienet-sdk just to run infra part as an alternative to ‎zombienet.toml? Do we really need to rewrite the PAPI tests to the rust? Why? When we have PET tests, I am sure that we want to rewrite them to the zombienet-sdk.
> 
> My understanding of #28 "Rewrite actual zombienet setup to use zombienet-sdk", was really just to rewrite zombient.toml to zombienet-sdk, but just the network setup and not all the tests. Am I missing anything here?

I'm not sure it's better to have rust tests, I just assumed that's what you meant for this issue because sdk is a rust library (it has a cli as well which should be more or less compatible with current toml). Alternatively, we can keep current toml(possibly slightly modified) and change the binary or write equivalent network config in rust(but not rewrite the tests).

bkontur:
>  write equivalent network config in rust(but not rewrite the tests).

Well, I wanted to start with this exactly :)

@pepoviola I thought you meant just this, replace zombienet.toml with zombienet-sdk network setup, right?

RafalMirowski1:
> > write equivalent network config in rust(but not rewrite the tests).
> 
> Well, I wanted to start with this exactly :)
> 
> @pepoviola I thought you meant just this, replace zombienet.toml with zombienet-sdk network setup, right?

ok, can do, in that case i'll close this pr and open another one

karolk91:
@bkontur, @RafalMirowski1 , @pepoviola

I see zombienet-sdk as an orchestration software, so you can imagine zombienet-sdk program to be responsible for starting the network, and then executing some tests. These tests can also mean "run typescript program" or use some other rust code to execute implement test scenarios

As an orchestartor zombienet-sdk provides very nice features to monitor the network for its state:
- did the parachain started to produce blocks already? if yes, then execute some tests
- do we expect something to be displayed in the logs ? zombienet-sdk has utils to get this
- do we need to some advanced stuff, change config, restart nodes then execute tests? zombienet-sdk can do it
- do we need to override some genesis stuff before running network ? zombienet-sdk can do it and then start the network

so we are basically avoiding implementing all the "do something before start, wait for node running, rpc responding, block number X and whatever" something we often cover with `just` recipes - its provided by zombienet-sdk, written in rust, can parallelize stuff if wanted easily

PET tests is totally different beast - its using chopsticsk to fork from already running network. So we can imagine Zombienet starting the network from scratch (wait for everything included) and then rust program can execute PET to do it stuff. But PET are often used to run tests agianst live network state - not the empty state started from scratch by zombienet

PAPI scripts - does it matter if we run them from bash? or rust program (zombienet-sdk) execute another program ?

We can argue if we should implement "business logic of tests" using rust or typescript but I think we eventually need both, while we need PAPI more - since PAPI tests we can execute within zombienet-sdk environment, but we can also re-use these to run against live networks

I wonder how @pepoviola see this architecture

pepoviola:
Hi @bkontur / @karolk91 / @RafalMirowski1, I think the first step can be just use the `zombie-cli` (from the rust version) instead of the typescript one (since our idea is to deprecate that version and all new features and improvements will go into the rust version). The `zombie-cli` have a _compatible_ interface and you can spawn the network with the same `toml` definitions and you also get the `zombie.json` with all the info.

Then related to the testing logic, both _papi_ or _zombienet-sdk_ have different ways to interact with the running network and using one or another is just a test implementation decision (but again both are good options :))

> PET tests is totally different beast - its using chopsticsk to fork from already running network. So we can imagine Zombienet starting the network from scratch (wait for everything included) and then rust program can execute PET to do it stuff. But PET are often used to run tests agianst live network state - not the empty state started from scratch by zombienet

For `PET` tests we use `zombie-bite` in AHM in order to fork a live network with the _current_ state with zombienet, if you need this kind of test I can help to implement this in CI.

Ping me if you have questions, I can also help with this transition to use the `zombie-cli` :)

karolk91:
> Hi @bkontur / @karolk91 / @RafalMirowski1, I think the first step can be just use the `zombie-cli` (from the rust version) instead of the typescript one (since our idea is to deprecate that version and all new features and improvements will go into the rust version). The `zombie-cli` have a _compatible_ interface and you can spawn the network with the same `toml` definitions and you also get the `zombie.json` with all the info.
> 
> Then related to the testing logic, both _papi_ or _zombienet-sdk_ have different ways to interact with the running network and using one or another is just a test implementation decision (but again both are good options :))
> 
> > PET tests is totally different beast - its using chopsticsk to fork from already running network. So we can imagine Zombienet starting the network from scratch (wait for everything included) and then rust program can execute PET to do it stuff. But PET are often used to run tests agianst live network state - not the empty state started from scratch by zombienet
> 
> For `PET` tests we use `zombie-bite` in AHM in order to fork a live network with the _current_ state with zombienet, if you need this kind of test I can help to implement this in CI.
> 
> Ping me if you have questions, I can also help with this transition to use the `zombie-cli` :)

Thanks @pepoviola

But specifically, how do you think zombienet-sdk fits with typescript-written tests ? I mean, using zombienet-sdk to spin the network, do some early checks like chain is up and running, blocks are being produced and then run typescript script via some `run_script` ? Obviously the same logic of the "test" could be implemented in rust - but the point here being also to confirm that typescript programs work fine against new runtime etc

pepoviola:
Thanks @karolk91, related to

> But specifically, how do you think zombienet-sdk fits with typescript-written tests ? I mean, using zombienet-sdk to spin the network, do some early checks like chain is up and running, blocks are being produced and then run typescript script via some run_script ? Obviously the same logic of the "test" could be implemented in rust - but the point here being also to confirm that typescript programs work fine against new runtime etc

`zombienet-sdk` fits with _ts test_ by providing the `spawn` phase and give you all the info of the network in the zombie.json file. You can totally re-use all the `ts/js` logic you have and even you can attach the _old ts test-runner_ if needed.

bkontur:
> Ping me if you have questions, I can also help with this transition to use the `zombie-cli` :)

@pepoviola iiuc, we should use zombie-cli instead of zombienet binary for the beginning? Because zombienet binary won't be supported soon?

Issue #35, opened 2026-03-05
[ProviderNode] Refactor CLI configuration


bkontur:
## TODO

- [ ] clean to cli.rs
- [ ] align with Polkadot-SDK style
- [ ] use clap args for everything (except secrets)
  - [ ] `--keyfile  /path/to/secret` with restricted access

ilchu:
Fixed in #38.

PR #38, opened 2026-03-09
Refactor cli config

RafalMirowski1:
https://github.com/paritytech/web3-storage/issues/35
- args from `main.rs`turned into clap args (but kept env var fallback)
- `--keyfile` with proper permissions can be used for seed or `--dev` flag which sets it to `"//Alice"`
- parsing of cli args is now done in `cli.rs`
- logic to run the node moved to `command.rs`


RafalMirowski1:
hey @bkontur is this roughly similar to what you had in mind? i wasn't exactly sure what Polkadot-SDK style means here

.github/workflows/integration-tests.yml
```diff
@@ -127,8 +127,10 @@ jobs:
 
       - name: Start provider (inmemory) and wait for health
         run: |
-          SEED="//Alice" CHAIN_RPC="ws://127.0.0.1:2222" BIND_ADDR="0.0.0.0:3333" \
-            nohup ./target/release/storage-provider-node --storage-mode inmemory > /tmp/provider.log 2>&1 &
+          nohup ./target/release/storage-provider-node \
+            --dev --storage-mode inmemory \
```
bkontur: @RafalMirowski1 let's remove `--dev` and do the same here as for bob with key file

RafalMirowski1: removed `--dev` flag entirely (or you meant just remove it here but keep as a flag?)

provider-node/src/cli.rs
```diff
@@ -0,0 +1,320 @@
...
+#[derive(Debug, clap::Args)]
+pub struct ReplicaSyncParams {
+    /// Enable autonomous replica sync.
+    #[arg(long, env = "ENABLE_REPLICA_SYNC")]
+    pub enable_replica_sync: bool,
+
+    /// Seconds between replica sync poll checks.
+    #[arg(
+        long,
+        value_name = "SECS",
```
franciscoaguirre: Can we not use an abbreviation here?


```suggestion
        value_name = "SECONDS",
```

provider-node/src/cli.rs
```diff
@@ -0,0 +1,320 @@
...
+        default_value_t = 12,
+        env = "REPLICA_POLL_INTERVAL"
+    )]
+    pub replica_poll_interval: u64,
+
+    /// Seconds before a replica sync operation times out.
+    #[arg(
+        long,
+        value_name = "SECS",
```
franciscoaguirre: Same

franciscoaguirre (approved):

provider-node/src/cli.rs
RafalMirowski1: done

provider-node/src/cli.rs
RafalMirowski1: done

provider-node/src/command.rs
```diff
@@ -0,0 +1,142 @@
...
+            let state = ProviderState::with_seed(storage, &seed)?;
+            tracing::info!("Signing enabled for account: {}", state.provider_id);
+            Arc::new(state)
+        }
+        None => {
+            let provider_id = cli
+                .key
+                .provider_id
+                .clone()
+                .unwrap_or_else(|| DEFAULT_PROVIDER_ID.to_string());
```
bkontur, line 53: @RafalMirowski1 do we need this DEFAULT provider? Can we just throw error? Does it make sense to run provider withtou provider id?

bkontur (approved):

provider-node/src/command.rs
bkontur: @RafalMirowski1 please follow up if needed :)

PR #39, opened 2026-03-11
Change zombienet binary

RafalMirowski1:
https://github.com/paritytech/web3-storage/issues/28

Changed zombienet binary, encountered issue where relative paths in zombienet.toml were not working so also intorduced `PROJECT_ROOT` env variable which is used in `zombienet.toml` like `default_command = "{{PROJECT_ROOT}}/.bin/polkadot"`


RafalMirowski1:
It looks like new zombie-cli binary is not straight-forwardly compatible with being run inside `paritytech/ci-unified:bullseye`. It fails with missing `libssl3`: https://github.com/paritytech/web3-storage/actions/runs/22950511272/job/66613712436 and even when we add `libssl3`, `libc` is incompatible: https://github.com/paritytech/web3-storage/actions/runs/22950984655/job/66615310176. Running on `ubuntu-latest` works, but not sure we wanna do this.
Maybe we should ask for a compatible binary to be built? @bkontur @karolk91 wdyt?

karolk91:
@RafalMirowski1, I think you could separate the process of building whatever is necessary (this would use the ci-unified image), and once you have all the binaries built (or maybe downloaded from cache) you could then have a separate step (which uses ubuntu image) to run the zombienet network and tests

I guess the questions are:
- why do we need to use ci-unified image? - IMHO, if we need to build any rust project, its good to re-use the image that polkadot-sdk uses for runtime builds reproducibility etc
- do we need to use ci-unified image for non-rust related stuff? - I think we dont need to
- can we run ubuntu images on parity runners? - I think we can find examples of this in the wild so I guess its not an issue

karolk91:
@RafalMirowski1, even though my previous comment could unblock us now, I think your idea
> Maybe we should ask for a compatible binary to be built?

is good in my opinion - I guess why we want to produce zombienet binaries that don't follow the same practices as `polkadot`, `polkadot-parachain` or other binaries. I think it would be good idea that zombienet sdk project keeps consistency with these. So definitely go and propose that in the zombienet project

RafalMirowski1:
> @RafalMirowski1, even though my previous comment could unblock us now, I think your idea
> 
> > Maybe we should ask for a compatible binary to be built?
> 
> is good in my opinion - I guess why we want to produce zombienet binaries that don't follow the same practices as `polkadot`, `polkadot-parachain` or other binaries. I think it would be good idea that zombienet sdk project keeps consistency with these. So definitely go and propose that in the zombienet project

For reference, requested here: https://github.com/paritytech/zombienet-sdk/issues/522

pepoviola:
_small update_: Latest version of zombie-cli [v0.4.8](https://github.com/paritytech/zombienet-sdk/releases/tag/v0.4.8) includes a `musl` version that is compatible with `ci-unified`.

Thx!

PR #52, opened 2026-04-28
feat: create web3-storage-paseo runtime

danielbui12:
## Issue

https://github.com/paritytech/web3-storage/issues/54


runtimes/web3-storage-paseo/src/genesis_config_presets.rs
```diff
@@ -0,0 +1,122 @@
...
+
+use crate::*;
+use alloc::{vec, vec::Vec};
+use cumulus_primitives_core::ParaId;
+use frame_support::build_struct_json_patch;
+use sp_genesis_builder::PresetId;
+use sp_keyring::Sr25519Keyring;
+
+// TODO(tungbui): Update Parachain id for Paseo Web3 Storage runtime
+pub const PARA_ID: ParaId = ParaId::new(4000);
```
danielbui12: How do we determine the `PARA_ID`?

runtimes/web3-storage-paseo/src/lib.rs
```diff
@@ -0,0 +1,807 @@
...
+        .for_class(DispatchClass::Operational, |weights| {
+            weights.max_total = Some(MAXIMUM_BLOCK_WEIGHT);
+            weights.reserved = Some(
+                MAXIMUM_BLOCK_WEIGHT - NORMAL_DISPATCH_RATIO * MAXIMUM_BLOCK_WEIGHT
+            );
+        })
+        .avg_block_initialization(AVERAGE_ON_INITIALIZE_RATIO)
+        .build_or_panic();
+    // TODO(tungbui): Update SS58 prefix for Paseo Web3 Storage runtime.
+    pub const SS58Prefix: u16 = 42;
```
danielbui12: What should `SS58Prefix` be?

scripts/build-paseo-chain-spec.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+# Build runtime and generate chain spec for zombienet
+set -e
+
+cd "$(dirname "$0")/.."
+
+# Clean up any existing chain spec
+rm -f chain_spec.json
```
ilchu, line 8: I'm not sure removing a preexisting chain spec silently would be an expected behaviour.

scripts/build-paseo-chain-spec.sh
```diff
@@ -0,0 +1,25 @@
...
+  -i "storage-paseo" \
+  -t local \
+  -p 1502 \
+  -c westend-local \
+  -r target/release/wbuild/storage-paseo-runtime/storage_paseo_runtime.compact.compressed.wasm \
+  named-preset local_testnet
+
+# Output the generated chain spec and clean up
+cat chain_spec.json
+rm chain_spec.json
```
ilchu, line 25: Maybe we could do it like bulletin? Generate a file and just log the location it was saved to, like `zombienet/chain_spec.json` or something.

runtimes/web3-storage-paseo/src/genesis_config_presets.rs
```diff
@@ -0,0 +1,122 @@
...
+            vec![
+                (
+                    Sr25519Keyring::Alice.to_account_id(),
+                    Sr25519Keyring::Alice.public().into(),
+                ),
+                (
+                    Sr25519Keyring::Bob.to_account_id(),
+                    Sr25519Keyring::Bob.public().into(),
+                ),
+            ],
```
ilchu, line 70: The UIs at least had 5 accounts preset (Alice, Bob, Charlie, David and Eve) or so. We might want to test primary and secondary provider setups, so we'd have at least two accounts for that. Those are technically offchain though, and I'm not sure if they'd need to be in the collators list.

scripts/build-paseo-chain-spec.sh
danielbui12: From this config in zombienet, it will run this script every time instead of using a generated file  https://github.com/paritytech/web3-storage/blob/fd08e16a2a86f3558c5279b6a77cf7e61a4133aa/zombienet/storage-paseo-local.toml#L23

Zombienet will read the chain spec from stderr then

scripts/build-paseo-chain-spec.sh
danielbui12: > From this config in zombienet, it will run this script every time instead of using a generated file  https://github.com/paritytech/web3-storage/blob/fd08e16a2a86f3558c5279b6a77cf7e61a4133aa/zombienet/storage-paseo-local.toml#L23
>
> Zombienet will read the chain spec from stderr then

I think for Paseo runtime after deployed we store the chain spec; but for development; I would prefer this way because the runtime is unstable

runtimes/web3-storage-paseo/src/genesis_config_presets.rs
danielbui12: from this code in pallet-storage-provider, everyone can stake then become storage provider (without collator role) https://github.com/paritytech/web3-storage/blob/be21a0fc831374fa710646b29c823fb406a36e41/pallet/src/lib.rs#L808-L869

danielbui12:
Move to https://github.com/paritytech/web3-storage/issues/58

PR #66, opened 2026-05-05
UI smoke tests, unit tests in CI

ilchu:
A small workspace package that hosts the Playwright fixture + waiters shared across the three UIs. Patterned after console-ui's e2e/fixtures in polkadot-bulletin-chain.

- `makeLocalPageFixture(opts)`: returns a Playwright `test` extended with a `localPage` fixture that pre-injects per-UI localStorage before navigation, then waits for chain connection (block-number testid) and at least block No.3 to clear mortal-era issues.
- `waitForConnection(page), `waitForMinBlock(page, n)`: standalone waiters for tests that need them outside the fixture.
- `probeProviderHealth()`: optional pre-test probe so tests fail with a clear hint rather than a 30s timeout when the provider node isn't running.

Also wires top-level test-orchestration scripts into `user-interfaces/package.json` so `npm run test:unit` etc. forwards to the right workspace.


.github/workflows/ui-checks.yml
```diff
@@ -0,0 +1,119 @@
...
+      - uses: actions/cache@v4
+        with:
+          path: ${{ steps.pnpm-store.outputs.STORE_PATH }}
+          key: pnpm-store-${{ runner.os }}-${{ hashFiles('user-interfaces/**/package.json', 'user-interfaces/pnpm-workspace.yaml') }}
+          restore-keys: |
+            pnpm-store-${{ runner.os }}-
+
+      - name: Install workspace deps
+        working-directory: user-interfaces
+        run: pnpm install --frozen-lockfile=false
```
danielbui12: Why shouldn’t we lock the version of packages?

danielbui12 (approved):

.github/workflows/ui-checks.yml
ilchu: Thanks for catching, definitely better to pin.

PR #69, opened 2026-05-05
Make the S3 and FS CI tests order independent

### LINKED ISSUE #22: CI for File System
Take care of complete CI Job as part for File System / S3

danielbui12:
### Issue 

Closes https://github.com/paritytech/web3-storage/issues/22.

### Summary:
- This PR resolve edge case of `s3-demo-ci` and `fs-demo-ci` execution
- Adopt the rest of `log`, `env_log` to `tracing`

---

The CI tests for `s3-demo-ci` and `fs-demo-ci` passed due to an edge case:
- It set up the node, provider
- Run `demo` command which executes `full-flow.js` (in this script already does e2e flow - that's why `s3-demo-ci` and `fs-demo-ci` passed)
https://github.com/paritytech/web3-storage/blob/df1eef5e5d1155ec8190bc489a3ca4ff68e39d35/.github/workflows/integration-tests.yml#L170-L180

If we run `s3-demo-ci` and `fs-demo-ci` command standalone, they will fail. So this PR aim to resolve that by adding provider registration step directly in `start-provider` command.

https://github.com/paritytech/web3-storage/blob/77e1b75eefc78f2edfa51b21d8fcef5be1505a60/justfile#L138-L140


client/examples/register_provider.rs
```diff
@@ -0,0 +1,111 @@
...
+
+    let mut provider_client = ProviderClient::new(config, ss58_address)?;
+    provider_client.connect().await?;
+    provider_client.set_signer(keypair.clone())?;
+
+    // Step 1: Register (idempotent — skip if already registered).
+    const STAKE: u128 = 1_000_000_000_000_000; // 1000 tokens (MinProviderStake, 12 decimals)
+
+    println!("Registering provider...");
+    #[allow(unused_assignments)]
```
ilchu: Why would we need to disable this lint?

client/examples/register_provider.rs
```diff
@@ -0,0 +1,111 @@
...
+        let uri: SecretUri = seed.trim().parse()?;
+        Keypair::from_uri(&uri)?
+    } else {
+        Keypair::from_uri(&"//Alice".parse::<SecretUri>()?)?
+    };
+
+    // Derive SS58 address from the keypair for display and ProviderClient identity.
+    let public_key_bytes = keypair.public_key().0;
+    let account = sp_runtime::AccountId32::from(public_key_bytes);
+    use sp_core::crypto::Ss58Codec;
```
ilchu: I'm surprised `cargo fmt` doesn't force it out of main into the use declarations.

storage-interfaces/file-system/client/src/lib.rs
```diff
@@ -879,53 +879,38 @@ impl FileSystemClient {
...
-            .map_err(|e| FsClientError::Blockchain(format!("Failed to submit tx: {e}")))?;
-
-        // Wait for finalization and extract drive_id from event
-        while let Some(event) = progress.next().await {
-            let event =
-                event.map_err(|e| FsClientError::Blockchain(format!("Transaction error: {e}")))?;
-
-            if let Some(finalized) = event.as_finalized() {
-                // Fetch events from the finalized block
-                let events = finalized.fetch_events().await.map_err(|e| {
```
ilchu, line 897: I appreciate the brevity, but why do we need the change overall? Was it failing weirdly if we tried to finalize per-event rather than per-tx?

client/examples/register_provider.rs
ilchu, line 1: Something about running an example file routinely as part of some other bigger and more common action just doesn't sit right with me. They are the first ones to experience code drift and get forgotten IMO. I'd say for sign and register functionality it's enough to just have it in the client. Or e.g. the change could be to make it part of provider setup at all times, or the example could be its own proper binary. I'm fine with either.

ilchu (commented):
Could be a decent QoL improvement, but tbh I couldn't immediately see neither how it relates to tests nor to CI (as it should according to the PR title). The CI is already testing the existing workflows and it all comes out green AFAICT (re: the issue you link), so I don't grasp the stimulus for making these changes besides starting work on client SDK -- which is a great thing to do, but I'd love to have it reflected in the title. It would save me and other reviewers some time if you could make it a bit more semantic and descriptive, both at the PR description and commit message level.

client/examples/register_provider.rs
danielbui12: closer look at the match block below
```rust
match provider_client
        .register(multiaddr, public_key_bytes.to_vec(), STAKE)
        .await
    {
        Ok(()) => {
            println!("  Provider registered");
            is_continue = true;
        }
        Err(e) if e.to_string().contains("ProviderAlreadyRegistered") => {
            println!("  Provider already registered");
            is_continue = false;
        }
        Err(e) => return Err(e.into()),
    }
```

In `Err()` case that `is_continue` is not even used. so the compiler warned

storage-interfaces/file-system/client/src/lib.rs
danielbui12: The previous approach is a bit weird for me;
It did:
1. submit tx
2. blindly wait for event
2.1. In case transaction failure, we cannot even see the error message reverted from pallet 

In stead, we have built-in method `wait_for_finalized_success` from subxt to ensure transaction complete. If tx failed, it would showed the error message.

Plus, take a look at `client`, most methods use `wait_for_finalized_success` as well

client/examples/register_provider.rs
danielbui12: Agree! this script is created for 1 time provider setup


https://github.com/paritytech/web3-storage/blob/77e1b75eefc78f2edfa51b21d8fcef5be1505a60/justfile#L118-L149

client/examples/register_provider.rs
danielbui12: anw, I updated the code to query if provider existed

danielbui12:
### Summary:
- This PR resolve edge case of `s3-demo-ci` and `fs-demo-ci` execution
- Adopt the rest of `log`, `env_log` to `tracing`

---

@ilchu 
> Could be a decent QoL improvement, but tbh I couldn't immediately see neither how it relates to tests nor to CI (as it should according to the PR title). The CI is already testing the existing workflows and it all comes out green AFAICT (re: the issue you link), so I don't grasp the stimulus for making these changes besides starting work on client SDK -- which is a great thing to do, but I'd love to have it reflected in the title. It would save me and other reviewers some time if you could make it a bit more semantic and descriptive, both at the PR description and commit message level.

The CI tests for `s3-demo-ci` and `fs-demo-ci` passed due to an edge case:
- It set up the node, provider
- Run `demo` command which executes `full-flow.js` (in this script already does e2e flow - that's why `s3-demo-ci` and `fs-demo-ci` passed)
https://github.com/paritytech/web3-storage/blob/df1eef5e5d1155ec8190bc489a3ca4ff68e39d35/.github/workflows/integration-tests.yml#L170-L180


If we run `s3-demo-ci` and `fs-demo-ci` command standalone, they will fail. So this PR aim to resolve that by adding provider registration step directly in `start-provider` command.

https://github.com/paritytech/web3-storage/blob/77e1b75eefc78f2edfa51b21d8fcef5be1505a60/justfile#L138-L140

ilchu (approved):
Great, now I finally understand the impetus.

PR #70, opened 2026-05-06
UI e2e suite and CI

ilchu:
Three threads landed together:

  1. **Tier-2 e2e suite (new).** Feature-level Playwright specs across all three UIs (44 tests total). Shared `@web3-storage/test-helpers` package: dev signers, chain-aware bucket/drive helpers, `cleanProviderRegistry` for deterministic cross-suite state. New
  `ui-e2e.yml` workflow gated from day 1 (no `continue-on-error`). Wall-clock: drive-ui 14.5m (19/19), console-ui 8.8m (14/14), provider 6.7m (11/11).

  2. **PAPI consolidation.**
     - Provider migrated to PAPI v2 single typed client (1223→611 LOC in `chain-client.ts`); eliminates a class of cross-client chainHead-lag bugs.
     - All 3 UIs drop `@polkadot/keyring|util|util-crypto` for `@polkadot-labs/hdkd` + `@polkadot-api/substrate-bindings`.
     - Single shared `@web3-storage/papi` workspace package owns the only tracked metadata snapshot. Per-UI `.papi/` directories deleted. `papi:check` only diffs `.papi/metadata` so zombienet's per-spawn genesis hash doesn't false-positive.

  3. **drive-ui catch-up to simplified runtime.** Specs trimmed to the slim `DriveInfo` (8 fields). Tests for removed extrinsics (`update_drive_name`, `clear_drive`, `commit_changes`, `update_root_cid`) wrapped with `test.fixme` referencing
  `user-interfaces/PAPI_OVERHAUL.md` for the source-code follow-up.

  ## Hardening worth calling out
  - `cleanup_bucket_internal` now drains pending `AgreementRequests` on bucket delete (avoids stuck requests when a delete races a provider's mid-flight accept).
  - drive-ui suite 24.7m → 14.5m at 19/19 green via shared drives in `members`/`persistence` specs and best-block cleanup submission (`submitExtrinsicBestBlock`).
  - console-ui suite stabilised by deriving "active account" from the global signer (was page-local state desync), an optimistic-insert in member-add (chainHead lag was clobbering the new row on refresh), and a UI-side duplicate-member guard (runtime updates the
  role rather than rejecting).


danielbui12 (commented):
@ilchu
- Please double check the CI  E2E test failed
- Should we put all `.papi` folders into `.gitignore` cause they are generated?

ilchu:
> @ilchu
> 
>     * Please double check the CI  E2E test failed
> 
>     * Should we put all `.papi` folders into `.gitignore` cause they are generated?

Sure, I'm working on it, the tests are trickier to setup than I thought. I'll convert it to draft until then. And PAPI stuff had some other issues too that I'll fix in a separate PR.

pallet/src/lib.rs
```diff
@@ -3468,6 +3468,36 @@ pub mod pallet {
...
+                    request.payment_locked
+                };
+                T::Currency::unreserve(&request.requester, total_locked);
+                Self::deposit_event(Event::AgreementRequestWithdrawn {
+                    bucket_id,
+                    provider,
+                    payment_returned: total_locked,
+                });
+            }
+
```
danielbui12, line 3500: Good catch!

However, the code iterates the entire `AgreementRequests` double-map (all providers × all buckets); the weight is too expensive.

Plus, changes to the function logic would require re-benchmarking the pallet as well. Since this PR aims to fix UI E2E and CI issues, I would not include this in the PR. 

Claude suggestions:
- Modify / create more storage interfaces to track the data better. (I think this is not a good choice, because change storage interfaces leads to storage migration, might be over engineering)
- Minimal solution: iterate `Providers` instead of `AgreementRequests`. `AgreementRequests` is unbounded over chain history (every successful accept removes one, but a chain that's seen N requests over its lifetime can still have N stale-ish entries from buckets that came and went without going through this cleanup path). `Providers` is a current-state set — every entry costs real stake to create, so its size is gated by economic admission, not by request churn.
```rust
for provider in Providers::<T>::iter_keys() {                                                                                                                     
      let Some(request) = AgreementRequests::<T>::take(&provider, bucket_id) else { continue };                                                                     
      // ... existing unreserve + event emission ...                                                                                                                
  }
```

PR #78, opened 2026-05-12
feat: add more papi examples & integration tests

danielbui12:
## Changes

- Update client & test about storage query regarding storage changes of
  - update `AgreementRequests` https://github.com/paritytech/web3-storage/pull/79
  - update `CheckpointRewards` & change provider deregister flow https://github.com/paritytech/web3-storage/pull/80
- Change runtime usage & add more config for provider node in CI integration tests
- Add more examples & let the CI integrations test verify them across all runtimes


danielbui12:
It seems the provider node does not work sustainably, 
- ProviderNotInSnapshot: https://github.com/paritytech/web3-storage/actions/runs/25731923320/job/75559086812
- Provider did not poll & accept agreement request: https://github.com/paritytech/web3-storage/actions/runs/25776488690/job/75710155576, https://github.com/paritytech/web3-storage/actions/runs/25790965497/job/75756140734?pr=78

Sometimes, does the node:
- AgreementNotFound: https://github.com/paritytech/web3-storage/actions/runs/25742570649/job/75596832304
- Did not receive BucketFrozen event: https://github.com/paritytech/web3-storage/actions/runs/25721859238/job/75544500820

=> Because the CI used the wrong runtime and provider node missed `--enable-agreement-coordinator` & `--enable-checkpoint-coordinator`

.github/workflows/integration-tests.yml
```diff
@@ -170,15 +170,36 @@ jobs:
       - name: Run L0 demo (inmemory provider)
         run: just demo "http://127.0.0.1:3333" "//Alice" "//Bob"
 
+      - name: Drain pool
+        run: bash scripts/drain-pool.sh ws://127.0.0.1:2222
+
```
bkontur: Can we rewrite scripts/drain-pool.sh to the justfile?
And trigger this drain-pool hash conditionally, something like:
```
run: just --drain-pool demo "http://127.0.0.1:3334" "//Charlie" "//Dave"
```

`--drain-pool` does not need to be specified, should be optinal

```suggestion
```

bkontur (changes requested):
Per https://github.com/paritytech/web3-storage/pull/78/files#r3240359470

.github/workflows/integration-tests.yml
danielbui12: just doesn't accept custom CLI flags, I'll do with `drain-then`
i.e `just drain-then demo "http://127.0.0.1:3334" "//Charlie" "//Dave"`

justfile
```diff
@@ -181,6 +181,52 @@ generate-chain-spec: build-runtime
...
+        sleep "$SLEEP_BETWEEN"
+    done
+    echo "drain-pool: WARNING - pool not empty after $((MAX_ITERS * SLEEP_BETWEEN))s, proceeding anyway"
+    sleep "$FINAL_BUFFER"
+
+# Drain the parachain pool, then run another recipe. Use in CI between
+# back-to-back integration tests to avoid stale-nonce drops.
+# Usage: just drain-then demo "http://127.0.0.1:3334" "//Charlie" "//Dave"
+[positional-arguments]
+drain-then RECIPE *ARGS: drain-pool
```
bkontur: I meant something like:
```
just polkadot_version=${{ env.POLKADOT_SDK_VERSION }} download-polkadot-sdk-binaries
```
but this is also interesting feature:

```suggestion
drain-tx-pool-then RECIPE *ARGS: drain-pool
```

.github/workflows/integration-tests.yml
```diff
@@ -85,11 +85,11 @@ jobs:
           node-version: "24"
 
       - name: Build runtime and provider
-        run: just build-runtime && just build-provider
+        run: just build-paseo-runtime && just build-provider
```
bkontur: @danielbui12 so now we are running the integration tests just for paseo runtime? I think we should run for both, testnet runtime and paseo runtime - matrix like style with binary caching, can you please take a look and create PR on the top of this PR?

danielbui12: @bkontur Created new PR https://github.com/paritytech/web3-storage/pull/82

.github/workflows/integration-tests.yml
```diff
@@ -199,3 +221,15 @@ jobs:
             /tmp/zombie-*/**/*.log
             /tmp/zombie-*/*-plain.json
             /tmp/zombie-*/*-raw.json
+
+  integration-tests-complete:
+    name: Integration Tests
```
bkontur, line 226: @danielbui12 yes, nice, those names are used by github config for required jobs to pass :)

bkontur (approved):

bkontur:
@ilchu please, also check

user-interfaces/console-ui/playwright.config.ts
```diff
@@ -5,7 +5,6 @@ export default defineConfig({
   globalSetup: "./e2e/global-setup.ts",
   fullyParallel: false,
   forbidOnly: !!process.env.CI,
-  retries: 0,
```
ilchu, line 8: I'd rather not drop these despite the default being same. Being explicit doesn't hurt.

examples/papi/api.js
ilchu, line 1: I feel this would almost qualify as a TS SDK. Considering what @bkontur said about not actually needing the rust side, good to see that just a few thin convenience wrappers get most of the way there.

justfile
```diff
@@ -181,6 +181,52 @@ generate-chain-spec: build-runtime
 demo PROVIDER_URL=PROVIDER_URL PROVIDER_SEED="//Alice" CLIENT_SEED="//Bob": papi-setup
     node examples/papi/full-flow.js "{{ CHAIN_WS }}" "{{ PROVIDER_URL }}" "{{ PROVIDER_SEED }}" "{{ CLIENT_SEED }}"
 
+# Wait until the parachain's transaction pool is empty (bounded ~60s, then
+# proceeds with a warning). Run between back-to-back integration tests so the
+# next step doesn't pick up an `accountNextIndex` that misses an in-flight tx
+# (which would land with the same nonce and get dropped as "Usurped").
```
ilchu, line 187: I had quite a few other nonce issues while refactoring the UIs, but they just keep coming. What a pain.

examples/papi/common.js
```diff
@@ -0,0 +1,602 @@
...
+ * Shared helpers for the PAPI examples in this directory.
+ *
+ * Importable from any example via `import { ... } from "./common.js"`.
+ */
+
+import { createClient } from "polkadot-api";
+import { getWsProvider } from "polkadot-api/ws-provider";
+import { getPolkadotSigner } from "polkadot-api/signer";
+import { Keyring } from "@polkadot/keyring";
+import { cryptoWaitReady, decodeAddress } from "@polkadot/util-crypto";
```
ilchu: Same comment here re: pdjs vs PAPI. It would be best to encode it as a Claude skill as he keeps trying to do it the old way.

examples/papi/bucket-membership.js
```diff
@@ -0,0 +1,101 @@
...
+ *
+ * Prerequisites:
+ *   - Parachain running at ws://127.0.0.1:2222
+ *   - Descriptors generated: npm run papi:generate
+ *
+ * Usage: node bucket-membership.js [chain_ws] [admin_seed] [writer_seed] [reader_seed]
+ */
+
+import assert from "node:assert";
+import { cryptoWaitReady } from "@polkadot/util-crypto";
```
ilchu: I'm surprised we still have the old pdjs somewhere, I thought I pruned all of it. And AFAICT this wait call is not needed overall with PAPI as it loads WASM synchronously and everything should be immediately callable at execution site.

examples/papi/common.js
```diff
@@ -0,0 +1,602 @@
...
+
+export function fmtRole(role) {
+  if (!role) return "?";
+  if (typeof role === "string") return role;
+  return role.type ?? JSON.stringify(role);
+}
+
+export function bytesToUtf8(maybeBin) {
+  if (!maybeBin) return "";
+  const bytes = maybeBin.asBytes ? maybeBin.asBytes() : maybeBin;
```
ilchu: From what I remember, `asBytes` is a purely PAPIv1 thing, and I transitioned all the UIs to v2. I believe we should do the same in examples and just don't put fallback shims at all. So overall this function could collapse to `return maybeBin ? new TextDecoder().decode(maybeBin) : ""` and maybe doesn't need to be a fn at all.

examples/papi/common.js
```diff
@@ -0,0 +1,602 @@
...
+    accepting_extensions: true,
+    max_capacity: 0n,
+  });
+}
+
+function bytesEq(a, b) {
+  if (!a || !b || a.length !== b.length) return false;
+  for (let i = 0; i < a.length; i++) if (a[i] !== b[i]) return false;
+  return true;
+}
```
ilchu, line 444: I'd rather not have abstraction for a single-time call.

examples/papi/common.js
```diff
@@ -0,0 +1,602 @@
...
+  return role.type ?? JSON.stringify(role);
+}
+
+export function bytesToUtf8(maybeBin) {
+  if (!maybeBin) return "";
+  const bytes = maybeBin.asBytes ? maybeBin.asBytes() : maybeBin;
+  return new TextDecoder().decode(bytes);
+}
+
+export function utf8(s) {
```
ilchu: Something like `toUTF8Bytes` would sound clearer. Also just like above: IMHO 1 line abstractions are noisy and little use.

examples/papi/full-flow.js
```diff
@@ -158,195 +59,83 @@ async function createAgreement(api, provider, client, bucketId) {
...
-  await api.tx.StorageProvider.accept_agreement({
-    bucket_id: bucketId,
-  }).signAndSubmit(provider.signer);
-  console.log("  Agreement accepted");
-}
-
-async function uploadData(api, bucketId) {
-  const data = new TextEncoder().encode(
-    `Hello, Web3 Storage! [${new Date().toISOString()}] provider=${PROVIDER_SEED}`
+  const maxBytes = 1_073_741_824n; // 1 GiB
```
ilchu, line 61: I had this bite me in the provider UI. Some places used 1 GiB (i.e. 1024 megabytes), and some 1 GB (i.e. 10^9 bytes). So this led to `InsufficientStakeForCapacity` revert when requesting. So I tried to align everything to use SI GB (10^9). Let's do the same here.

runtime/src/weights/xcm/pallet_xcm_benchmarks_fungible.rs
ilchu, line 1: Seems like it's the same naming issue as before. Also as a side note, wouldn't hurt to have the CI bot for benchmarks/formatting etc like you added in Bulletin. Just to have it done on single hardware.

ilchu (changes requested):
Some chore-like issues overall, but overall I appreciate greater example coverage, thank you.

runtime/src/weights/xcm/pallet_xcm_benchmarks_fungible.rs
bkontur: > Seems like it's the same naming issue as before. Also as a side note, wouldn't hurt to have the CI bot for benchmarks/formatting etc like you added in Bulletin. Just to have it done on single hardware.

yeah, no problem to add: https://github.com/paritytech/polkadot-bulletin-chain/pull/464

examples/papi/bucket-membership.js
danielbui12: Yep, I just knew about it recently

examples/papi/full-flow.js
danielbui12: @ilchu  but we should actually standardize to GiB, not GB. Polkadot SDK and HOP protocol in Bulletin chain handle memory strictly in IEC standard. Forcing SI standard leaves some measurement lost.

I still prefer GiB instead, let's figure out and fix them all.

bkontur: yes, let's be precise, GiB

ilchu: Gotcha, didn't look that up. Then I should revert back to GiBs in UI as well.

ilchu (approved):

.github/env
```diff
@@ -1,3 +1,5 @@
 RUST_STABLE_VERSION=1.88.0
```
bkontur:
```suggestion
RUST_STABLE_VERSION=1.93.0
```

.github/env
```diff
@@ -1,3 +1,5 @@
...
 POLKADOT_SDK_VERSION=polkadot-stable2603
 ZOMBIENET_VERSION=v0.4.11
+CI_IMAGE=paritytech/ci-unified:bullseye-1.88.0-2025-06-27-v202507112050
```
bkontur:
```suggestion
CI_IMAGE=docker.io/paritytech/ci-unified:bullseye-1.93.0-2026-01-27-v202603042356
```

PR #79, opened 2026-05-12
Flip AgreementRequests key order to improve performance of bucket cleanup

danielbui12:
## Context

The clean up  `AgreementRequests` iterates the whole storage double-map (all providers × all buckets); the weight is too expensive. 

## Changes

- Re-keys `AgreementRequests` from `StorageDoubleMap<AccountId, BucketId, _>` to `StorageDoubleMap<BucketId, AccountId, _>`.
- Replaces the full double map `iter().filter()` scan in `cleanup_bucket_internal` with `drain_prefix(bucket_id)`, dropping cleanup work from O(total pending requests on chain) to O(matched providers for this bucket).

Question (@bkontur): We has yet to deploy the runtime Paseo, dont we have to create storage migration for `pallet-storage-provider`?

## Issue

Address https://github.com/paritytech/web3-storage/pull/70#discussion_r3216172761


bkontur (approved):
no migration needed, nothing is deployed live/testnet.
In case of migration, you would need to change pallet's STORAGE_VERSION and migration, but not needed now

bkontur:
@danielbui12 please check  https://github.com/paritytech/web3-storage/actions/runs/25726767832/job/75541452776?pr=79

.github/workflows/ui-e2e.yml
```diff
@@ -170,10 +170,6 @@ jobs:
             sleep 2
           done
 
-      - name: PAPI descriptor drift check
```
danielbui12, line 173: cc @ilchu 
since we already put the `.papi` into `.gitignore` so we can remove this CI and `papi:check` script

PR #87, opened 2026-05-22
Update storage provider benchmark precisely

danielbui12:
## Chagnes

The previous one missed edge cases in detail. This PR aims to add more benchmarks to improve code coverage for these extrinsics:
- end_agreement
- respond_to_challenge

TODO: A sole admin of bucket can demote himself causing no admin left for a bucket. How should we handle this case?


pallet/src/lib.rs
danielbui12, line 1452: @bkontur should we enforce there is at least 1 admin for each bucket?

ilchu: I feel it should be done in a separate PR, but overall there's no benefit to allowing deleting **last** admin AFAICT. Like things mostly work, but a lot of functionality (bucket freezing, early termination, bucket membership changes,  responding to "deleted" challenges, deleting drives on L1...) becomes locked and not recoverable as there's no way for another provider to get promoted. So it's better to try to enforce there remaining at least one admin as an invariant.

ilchu (approved):
I appreciate being very particular about challenge response waits. My only comment is about your TODO, which is a nice issue to flag and fix, but maybe in a different PR.

PR #90, opened 2026-05-22
fix(ui): align byte literals with binary GB convention

ilchu:
## Summary

- The three UI `formatBytes()` helpers (`console-ui`, `drive-ui`, `provider`) already use base 1024 with `KB/MB/GB/TB/PB` labels — i.e. the colloquial "Windows-style" interpretation. But several hardcoded byte literals and SDK examples sat next to `// 1 GB` / `// 1 TB` comments while using SI values (`1_000_000_000n` = 10^9 etc.). The formatter would render those as "953 MB" / "931 GB", contradicting the comment.
- Switch the literals to the binary equivalents (`1_073_741_824n` = 2^30 for 1 GB, `1_099_511_627_776n` = 2^40 for 1 TB) so the comment matches what `formatBytes()` shows. Drop the misleading "matches runtime's SI convention" remark in `Registration.tsx` — the runtime stores raw byte counts and is convention-agnostic.
- `formatBytes` had zero unit-test coverage. Add `describe('formatBytes', …)` blocks in `provider/src/utils/format.test.ts` and a new `drive-ui/src/lib/__tests__/utils.test.ts`. They lock in `1 KB = 1024 B`, `1 GB = 2^30 B`, etc., with an explicit guard that 10^9 bytes is labelled `953.67 MB` (not "1 GB") so any future drift back to SI fails CI.

## Files

| File | Change |
|------|--------|
| `provider/src/pages/Registration.tsx` | `maxCapacity` default 10^12 → 2^40 |
| `console-ui/src/components/S3Tab.tsx` | "10 MB" default `"10000000"` → `"10485760"` |
| `sdk/typescript/{s3,file-system}/examples/basic-usage.ts` | `capacity` 10^9 → 2^30 |
| `sdk/typescript/file-system/src/client.ts` | JSDoc example 10^9 → 2^30 |
| `provider/src/utils/format.test.ts` | +`formatBytes` describe block (7 cases) |
| `drive-ui/src/lib/__tests__/utils.test.ts` | new file (7 cases) |

No formatter, no SDK type/runtime, and no docs/CLAUDE.md changes. Pre-existing typecheck and ESLint failures on `dev` are untouched.

### DIFF
--- a/user-interfaces/provider/src/utils/format.test.ts
+++ b/user-interfaces/provider/src/utils/format.test.ts
@@ -78,36 +78,37 @@ describe('formatTokens', () => {
 // Byte units use binary (base 1024) — colloquial "GB" == 2^30 bytes,
 // matching what most users mean when they type "1 GB".
 describe('formatBytes', () => {
-  it('formats zero and sub-KB values without a decimal', () => {
+  it('formats zero and sub-KB values', () => {
     expect(formatBytes(0)).toBe('0 B')
     expect(formatBytes(1)).toBe('1 B')
     expect(formatBytes(1023)).toBe('1023 B')
   })
 
   it('uses base 1024 for unit transitions', () => {
-    expect(formatBytes(1024)).toBe('1.00 KB')
-    expect(formatBytes(1024 ** 2)).toBe('1.00 MB')
-    expect(formatBytes(1024 ** 3)).toBe('1.00 GB')
-    expect(formatBytes(1024 ** 4)).toBe('1.00 TB')
+    expect(formatBytes(1024)).toBe('1 KB')
+    expect(formatBytes(1024 ** 2)).toBe('1 MB')
+    expect(formatBytes(1024 ** 3)).toBe('1 GB')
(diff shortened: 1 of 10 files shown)


user-interfaces/provider/src/utils/format.test.ts
```diff
@@ -74,3 +74,40 @@ describe('formatTokens', () => {
     expect(formatTokens(999n)).toBe('999 pico UNIT')
   })
 })
+
+// Byte units use binary (base 1024) — colloquial "GB" == 2^30 bytes,
+// matching what most users mean when they type "1 GB".
+describe('formatBytes', () => {
```
bkontur, line 80: here is similar: https://github.com/paritytech/polkadot-bulletin-chain/blob/65ecc2321f35fe19644b13a8518cfd1bafd504c2/console-ui/src/utils/format.ts#L1

bkontur (approved):

danielbui12 (approved):

PR #80, opened 2026-05-14
Provider lifecycle fixes

danielbui12:
## Changes
Resolve these issues:
- `update_provider_settings` allows shrinking `max_capacity` to barely-above-committed, but does not validate that existing agreements still satisfy other invariants; settings change is not gated by an event payload that includes the new values
- `add_stake` reserves currency before mutating provider stake; on `checked_add` overflow the reserved funds remain locked
- `set_extensions_blocked` allows a provider to block extensions on an agreement for any `bucket_id` they have — but does not validate the caller is still a registered provider with that agreement; relies on `AgreementNotFound` for indirect protection
<details>
  <summary>[Breaking Changes] <code>deregister_provider</code> ignores pending checkpoint rewards, unclaimed pool balance, and outstanding challenges</summary>

**Solution**

  Two-step exit modeled:
  1) Step: Announce
  - Extrinsic: deregister_provider (same call_index 2)
  - Effect: Forces `accepting_primary = false`, `accepting_extensions = false`. Stamps `deregister_at = System::block_number() + DeregisterAnnouncementPeriod`. Stake stays reserved; provider remains slashable.

  2) Step: Complete
  - Extrinsic: complete_deregister (new, call_index 6)
  - Effect: After period elapses: drains `CheckpointRewards` via `iter_prefix(&who)` into free balance, unreserves stake, removes provider.

  3) Step: Cancel
  - Extrinsic: cancel_deregister (new, call_index 7)
  - Effect: Clears deregister_at. Provider re-enables flags via update_provider_settings.

**Changes**:
    - `DeregisterAnnouncementPeriod` config constant (is set `ChallengeTimeout = 48 HOURS`). The `period >= ChallengeTimeout` guarantees every pre-announce challenge matures while the provider is still slashable.
    - `CheckpointRewards` double map storage layout flipped to `(AccountId, BucketId)` so `iter_prefix(&provider)` enables the drain.
    - `update_provider_settings` blocked when `deregister_at.is_some()` so the freeze can't be undone mid-window.
    - New errors `DeregisterAnnounced`, `DeregisterNotAnnounced`,  `DeregisterPeriodNotElapsed`, new events `DeregisterAnnounced`, `DeregisterCancelled`.

</details>

PR #96, opened 2026-05-27
Fix e2e test flakiness

ilchu:
## Summary

  Diagnosed and fixed the provider e2e flake (`registration.spec.ts:23 › fresh registration with Ferdie via wizard`) and two related state-leakage bugs found while auditing the broader e2e suites.

  The reported CI failure was the wizard's 120s `provider-registered-badge` wait timing out. The wizard submits `register_provider` + `update_provider_settings` sequentially and awaits finalization on each, then runs `loadProviderData` before the badge unblocks —
   locally that lands at ~72s, on `parity-large` it has overshot 120s.

  While reproducing locally I also hit a second mode: once any wizard run succeeds, Ferdie's `Providers` entry persists. The runtime's deregister is two-step (announce now, `complete_deregister` only callable after `DeregisterAnnouncementPeriod` = 48h), so
  `cleanProviderRegistry` can't fully unregister between runs, and subsequent runs see the SettingsManager instead of the wizard. A third issue: `cleanupBuckets` silently leaks any bucket with `object_count > 0` because `delete_s3_bucket` requires the bucket be
  empty.

  ## Changes

  - **`cleanProviderRegistry` skips already-announced providers** (`813d59e`). Re-announcing throws `DeregisterAnnounced` from the runtime; without the guard, a second globalSetup against the same chain aborts every test in the suite.
  - **`cleanupBuckets` drains `Objects` before calling `delete_s3_bucket`** (`46b2b36`, `8370a9e`). Iterates `S3Registry.Objects.getEntries(bucketId)` and deletes each object via `delete_object_metadata`. The follow-up commit fixes a type-cast that hid a runtime
  bug — `keyArgs[1]` is `Uint8Array` directly per the PAPI descriptor, not a Binary wrapper.
  - **Wizard test skips with an explanatory message when Ferdie is already registered** (`ff38034`). Local re-runnability needs a fresh chain; until the underlying two-step-deregister limitation is addressed, skipping cleanly beats failing.
  - **Wizard badge timeout 120s → 150s** (`df9afd4`). Fits inside the existing 180s spec budget after ~30s of pre-submit wizard navigation; absorbs the worst case observed in CI.

  ## Not fixed

  The wizard test's "Ferdie persists across runs" issue is worked around (skip), not resolved. A real fix needs either reducing `DeregisterAnnouncementPeriod` in the local runtime (currently 48h, with a `>= ChallengeTimeout` invariant) or adding a test-only
  force-remove path. Both touch the runtime; happy to do as a follow-up if wanted.

### DIFF
--- a/user-interfaces/drive-ui/e2e/integration/members.spec.ts
+++ b/user-interfaces/drive-ui/e2e/integration/members.spec.ts
@@ -72,9 +72,13 @@ test("add Reader → list refreshes without manual click", async ({ localPage })
   await localPage.getByTestId("add-member-submit").click();
 
   // After tx settles, Charlie's row should appear without clicking refresh.
-  // signAndSubmit waits for finalization (~24s on local zombienet) and the
-  // refresh follows. 45s is comfortable headroom.
+  // signAndSubmit waits for finalization (~24-36s on local zombienet) and
+  // refresh follows. 45s used to be "comfortable headroom" but CI runs the
+  // drive-ui suite back-to-back with file-ops which holds the chain busy
+  // for several minutes, and the set_member finalize has overshot 45s
+  // there. Match the 90s convention used by realtime.spec.ts for similar
+  // post-tx UI assertions.
   await expect(localPage.getByTestId(`member-row-${Charlie.address}`)).toBeVisible({
-    timeout: 45_000,
+    timeout: 90_000,
   });
 });
--- a/user-interfaces/provider/e2e/integration/registration.spec.ts
+++ b/user-interfaces/provider/e2e/integration/registration.spec.ts
@@ -53,8 +53,16 @@ test("fresh registration with Ferdie via wizard", async ({ localPage }) => {
   // After registration, the wizard redirects to SettingsManager (there's no
   // dedicated "complete" step — the Registered badge in the settings header
   // is the only "you're registered" notification the UI exposes).
+  //
+  // Budget: `submitRegisterProvider` submits two sequential extrinsics
+  // (`register_provider` + `update_provider_settings`) and awaits
+  // finalization on each (~24–36s per tx on a busy parachain), then
+  // `loadProviderData` runs another batch of queries before SettingsManager
+  // unblocks. Locally that's ~72s; on the CI parity-large runner it has
+  // run past 120s. 150s sits comfortably inside the 180s spec budget while
+  // absorbing the worst case.
   await expect(localPage.getByTestId("provider-registered-badge")).toBeVisible({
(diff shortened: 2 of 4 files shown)


bkontur (approved):
@ilchu is this related to the @danielbui12 similar fix: https://github.com/paritytech/web3-storage/pull/78/files#diff-4761c5069b46ce96f49d74476da61f43ced3abc73d22254cd95a2be0e250a593R190 ?

danielbui12:
> @ilchu is this related to the @danielbui12 similar fix: https://github.com/paritytech/web3-storage/pull/78/files#diff-4761c5069b46ce96f49d74476da61f43ced3abc73d22254cd95a2be0e250a593R190 ?

@bkontur  The problem is about provider registration flow was changed to 2-steps in https://github.com/paritytech/web3-storage/pull/80.

With first step is announce on-chain (to stop accept request, challenge, etc), second step is after `DeregisterAnnouncementPeriod = 48h` (in pallet Config), provider can choose either leave or stay. 
The `DeregisterAnnouncementPeriod` is blocker that E2E UI tests are not able to complete the flow.

danielbui12:
> The wizard test's "Ferdie persists across runs" issue is worked around (skip), not resolved. A real fix needs either reducing DeregisterAnnouncementPeriod in the local runtime (currently 48h, with a >= ChallengeTimeout invariant) or adding a test-only force-remove path

@ilchu For the UI E2E, we can ignore the `complete_deregister` / `cancel_deregister`; in integration tests of runtime in `runtimes/web3-storage-paseo/tests/tests.rs` I added tests to cover these cases already. @bkontur  wdyt?

ilchu:
> > The wizard test's "Ferdie persists across runs" issue is worked around (skip), not resolved. A real fix needs either reducing DeregisterAnnouncementPeriod in the local runtime (currently 48h, with a >= ChallengeTimeout invariant) or adding a test-only force-remove path
> 
> @ilchu For the UI E2E, we can ignore the `complete_deregister` / `cancel_deregister`; in integration tests of runtime in `runtimes/web3-storage-paseo/tests/tests.rs` I added tests to cover these cases already. @bkontur wdyt?

Fair point for the runtime, and good to know it's covered. Here though the problem was for local reruns on the same global setup: the account used for acknowledging the UI state didn't completely deregister, so it never saw the right screen. But I guess instead of skipping that particular test if the Ferdie account is already there like I did initially I'll try sudo removing the thing to make tests more reproducible. Not that it's a CI problem, but still.

danielbui12:

PR #95, opened 2026-05-27
Smart contract functionality via `revive` pallet

ilchu:
Closes #83.

  Wires pallet_revive 0.13.0 into both runtimes and exposes the client-side bucket lifecycle + drive registry to Solidity contracts via two custom precompiles.

  ## What's in

  - **Runtime** — pallet_revive at index 60 in storage-parachain-runtime and storage-paseo-runtime. UncheckedExtrinsic, TxExtension, WeightToFee, and runtime APIs all swapped to revive's Ethereum-flavored wrappers (asset-hub-westend pattern). Distinct ChainIDs
  420_420_500 / 420_420_501. AccountId32Mapper as the H160 ↔ AccountId32 bridge.
  - **Precompiles**
    - `0x…09010000` — storage-provider (11 selectors: createBucket, createBucketWithStorage, freezeBucket, set/removeMember, requestPrimaryAgreement, top/extend/endAgreement{Pay,Burn}, challengeCheckpoint)
    - `0x…09020000` — drive-registry (createDrive, deleteDrive, share/unshareDrive)
    - `0x…09030000` — s3-registry (createS3Bucket, createS3BucketWithStorage, deleteS3Bucket, putObjectMetadata, deleteObjectMetadata, copyObjectMetadata)
  - **Example dApps**
    - `StorageMarketplace.sol` — buys storage with `msg.value`, tracks per-user ownership, exposes `endMyAgreement` (storage-provider precompile).
    - `SharedTeamDrive.sol` — contract owns a drive, admin invites/kicks members, disbands on teardown (drive-registry precompile).
    - `TokenGatedDrive.sol` — contract owns an S3 bucket and mints a transferable NFT-shaped access token per stored object; transfer transfers access; burn deletes the object metadata (s3-registry precompile).
  - **Tests**
    - `just sc-demo` — full marketplace e2e (deploy → buyStorage → upload/challenge round-trip → endMyAgreement → provider paid)
    - `just sc-coverage` — direct precompile invocation for every selector across all three precompiles; on-chain event/state asserted per call
    - `just sc-team-drive` — SharedTeamDrive e2e (createTeam → invite → kick → disband)
    - `just sc-token-gated` — TokenGatedDrive e2e (initialize → mint → transfer → burn → shutdown)
  - **CI** — installs solc 0.8.35 + resolc 1.1.0 (pinned in `.github/env`), runs both demos across the matrix runtimes.
[...]

Issue #88, opened 2026-05-22
[Provider] Investigate provider validations (role, bucket, agrement, user, ...)


bkontur:
Relates to comment: https://github.com/paritytech/polkadot-sdk/pull/10731/files#r3242396564

iirc, the actual code turned off auth validation on provider.

# TODO

- [ ] Summarize
  - what validations is provider doing now
  - what validations are missing
    - cross-chain genesis hash validation?
  - what is the final set of validations on the provider

bkontur:
@danielbui12 please, start with summarize where are we now and what is missing.

cc: @eskimor

danielbui12:
The summary is in a flow-driven way. I believe it might be easier to read and verify

danielbui12:
## Flow 1: Data Upload and Commit

**Step 1.1 — PUT /node accepts a chunk or internal Merkle node**

- Current
  - Storage layer recomputes blake2_256(data) and rejects mismatch with InvalidHash (provider-node/src/storage/disk.rs:204-210).
  - For internal nodes, every non-zero child hash must already exist; missing ones return ChildrenMissing (provider-node/src/storage/disk.rs:213-235).
  - Per-hash dedup is enforced (provider-node/src/storage/disk.rs:258-263).
  - Bucket quota check exists (provider-node/src/storage/disk.rs:243-249) but is dormant because init_bucket is called with u64::MAX (provider-node/src/api.rs:210).
- Missing
  - MEDIUM - Bucket existence on-chain. The provider stores nodes for any bucket_id the caller supplies, without confirming the bucket exists on chain or that an agreement names this provider. Wastes disk and lets clients pollute namespaces, but not a direct slashing risk on its own.
  - MEDIUM - Quota wiring: init_bucket should derive max_bytes from the on-chain StorageAgreement.max_bytes. Today the quota check is dormant, so a misbehaving client can push the provider past committed_bytes and break the stake-to-bytes invariant.
  - MEDIUM - Refuse uploads when the bucket is frozen. Stored data is dead weight once the bucket is frozen at a prior seq.
  - MEDIUM - Refuse uploads when the agreement is past expiry or inside an expiry margin (margin should cover at least one checkpoint window so data has time to be anchored). Consumes the cached expires_at populated by AgreementAccepted / AgreementEnded events (see ## Flow 10).
[...]

PR #81, opened 2026-05-18
WIP: Event driven coordinators

bkontur:
---

_Closes #227 — cross-repo roadmap tracking._

PR #105, opened 2026-05-29
[BREAKING CHANGES] Move agreement negotiation off-chain, keep only mutual commitment on-chain

### LINKED ISSUE #97: Move agreement negotiation off-chain, keep only mutual commitment on-chain
## Summary

The current agreement flow puts negotiation logic on-chain that doesn't need to be there. The chain should only handle the mutual commitment that needs to be provable for disputes/slashing.

## Current Flow (3 extrinsics)

1. **`create_bucket_with_storage`** — Auto-discovers a provider on-chain via `find_matching_provider` (O(n) iteration over all providers), creates bucket + agreement without provider consent
2. **`request_primary_agreement`** — User submits agreement request on-chain, reserves funds, stores temporary `AgreementRequest`
3. **`accept_agreement`** — Provider accepts on-chain, finalizes the agreement

## Problems

- **`find_matching_provider` is O(n) on-chain work** — iterates all registered providers in runtime storage. This is a DoS vector as the provider set grows and wastes blockspace for logic that doesn't need a proof. https://hackmd.io/@C1JDzk4wSpu-_9KIRuGLgg/B1Wc1kNlzg
- **`request_primary_agreement` is just negotiation** — storing a temporary request on-chain burns gas and storage for coordination that can happen off-chain via HTTP.
- **`create_bucket_with_storage` bypasses provider consent** — it auto-selects and binds a provider without their explicit agreement to the specific deal.
- **Temporary storage overhead** — `AgreementRequests` storage map exists only to bridge two extrinsics, plus expiry/cleanup logic.

### Off-chain (client + provider HTTP)
1. Client discovers providers via runtime API queries (`query_find_matching_providers`, `query_available_providers`) — already implemented in `DiscoveryClient`
2. Client negotiates terms (price, duration, bytes) with provider via HTTP
3. Provider signs the agreed terms off-chain

### On-chain (single extrinsic)
4. **`establish_agreement(bucket_id, provider, terms, provider_signature)`** — verifies both parties' consent via signature, locks funds, creates the agreement

## Benefits

- **One extrinsic instead of two/three** for agreement creation
- **No O(n) on-chain iteration** — matching stays off-chain where it belongs
- **No temporary `AgreementRequest` storage** — no expiry cleanup needed
- **Provider consent is explicit** — provider signs the specific deal terms
- **Chain only does what needs a proof** — the binding commitment for disputes/slashing

## Scope

- [ ] Remove `create_bucket`, `create_bucket_with_storage` extrinsic
- [ ] Update replica storage request flow to use provider-signed agreement term
- [ ] Remove on-chain `find_matching_provider`
- [ ] Replace `request_primary_agreement` + `accept_agreement` with a single `establish_agreement` extrinsic that takes a provider signature
- [ ] Remove `AgreementRequests` storage map and related cleanup logic (`reject_agreement`, `withdraw_agreement_request`, request expiry)
- [ ] Add provider-side HTTP endpoint for agreement negotiation/signing
- [ ] Update `DiscoveryClient` / client SDK to use the new flow
- [ ] Update file system client (`storage-interfaces/file-system/client`) accordingly
- [ ] Update tests and benchmarks

danielbui12:
## Primary Changes

- `pallet-storage-provider` (Layer 0)
  - New `establish_storage_agreement(provider, terms, sig)` extrinsic — atomically verifies provider sig, checks replay window + expiry, creates bucket, opens agreement for primary provider.
  - New types: `AgreementTerms (owner, max_bytes, duration, price_per_byte, valid_until, nonce)`, `ProviderReplayStates (hwm + 256-bit bitmap per provider)`.
  - Sig scheme: `MultiSignature::verify(blake2_256(SCALE(terms)))` against the provider's registered public key.
  - Internal helper `establish_storage_agreement_internal` reused by `pallet-s3-registry` & `pallet-drive-registry`.
  - Removed: request_primary_agreement, request_replica_agreement, accept_agreement, withdraw/reject_agreement_request, create_bucket, create_bucket_with_storage, plus AgreementRequests
  storage + related events/errors.
  - New errors: InvalidProviderSignature, TermsExpired, NonceAlreadyUsed, NonceTooOld, TermsOwnerMismatch. New event: AgreementEstablished.

- Layer 1
  - `pallet-s3-registry::create_s3_bucket` and `pallet-drive-registry::create_drive` now take (terms, sig) and call to `pallet-storage-provider::establish_storage_agreement_internal`; removed old & unused code.

- client/ (Rust SDK)
  - Add new type `AgreementTermsOf`, and `sign_terms()` helper.
  - `AdminClient::establish_storage_agreement(provider, terms, sig)` replaces old flow.
  - `ProviderClient::negotiate_terms(url, req)` invokes POST `/negotiate` from `provider-node` to negotiate and get provider's signature on agreements.
  - Dropped all request/accept/withdraw/reject admin methods.

- provider-node/
  - New POST `/negotiate` → provider allocates nonce (per request), signs terms, returns `SignedTerms`.
  - Introduce `NonceCounter` that bootstraps from chain `hwm` on cold start. `NonceCounter` is used to prevent signature replay attack.
  - Dropped the old request/accept agreement coordinator.

## Issues
- Close: https://github.com/paritytech/web3-storage/issues/97

## Follow-up issue:
- https://github.com/paritytech/web3-storage/issues/124
- https://github.com/paritytech/web3-storage/issues/123


pallet/src/weights.rs
ilchu, line 1: As I've finally run the [bench](https://github.com/paritytech/web3-storage/pull/95#issuecomment-4591788615) the CI runner, no need to rerun it manually. Or if you need to regenerate things, still resort to the bot, it seems to run fine (though off dev by default).

primitives/src/agreement_term.rs
```diff
@@ -0,0 +1,53 @@
...
+//! [`AgreementTerms`] shape covers both flavours: `replica` is
+//! `None` for primary agreements and `Some(_)` for replica agreements,
+//! carrying the per-sync funding parameters.
+
+use codec::{Decode, DecodeWithMemTracking, Encode, MaxEncodedLen};
+use core::fmt::Debug;
+use scale_info::TypeInfo;
+
+/// Off-chain quote signed by the provider and redeemed on-chain by the owner.
+///
```
ilchu:
```suggestion

```

provider-node/src/api.rs
```diff
@@ -738,6 +741,37 @@ async fn get_historical_roots(
     }))
 }
 
+/// Negotiate provider-signed [`AgreementTerms`] for a bucket owner.
+///
+/// TODO: Request are automatically accepted, implement advance features to let providers determine
```
ilchu: Can you make an issue for this? And I feel that overall it would need to be wired into the provider dashboard as well. Especially considering that we have checkpoints and challenges exposed for manual resolution over there, and they are usually much more urgent than this.

provider-node/src/api.rs
```diff
@@ -738,6 +741,37 @@ async fn get_historical_roots(
...
+///
+/// Returns `503` if the node has no signing key (no `--keyfile`) or no
```
ilchu: This one got cut off, it seems.

provider-node/src/api.rs
```diff
@@ -738,6 +741,37 @@ async fn get_historical_roots(
...
+        Error::Internal("provider node has no nonce counter; /negotiate disabled".to_string())
+    })?;
+
+    let terms: AgreementTermsOf = AgreementTerms {
+        owner: req.owner,
+        max_bytes: req.max_bytes,
+        duration: req.duration,
+        price_per_byte: req.price_per_byte,
+        // TODO: lookup current block and use a bounded offset.
+        valid_until: u32::MAX,
```
ilchu: I would appreciate if it could land in this PR. And probably it should be a CLI flag or something similar for the provider binary. Or some more dynamic kind of knob with UI would be even better, but that's a bigger undertaking, of course.

provider-node/src/types.rs
```diff
@@ -245,8 +245,8 @@ pub struct ListBucketsResponse {
 /// Provider info response.
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct InfoResponse {
-    pub status: String,
-    pub version: String,
+    pub provider_id: String,
+    // TODO: Add more provider information
```
ilchu: Could be price per byte, total storage used, quote validity period etc.

primitives/src/agreement_term.rs
```diff
@@ -0,0 +1,53 @@
...
+    /// through the provider's sliding replay window.
+    pub nonce: u64,
+    /// Replica-specific parameters.
+    /// - `None` means these are primary terms;
+    /// - `Some(_)` means the provider has quoted a replica agreement and the extra per-sync funding is included.
+    pub replica_params: Option<ReplicaTerms<Balance, BlockNumber>>,
+}
+
+/// Replica terms
+///
```
ilchu:
```suggestion
/// Provider replica terms.
```
or something more verbose even.

primitives/src/provider_replay_state.rs
```diff
@@ -0,0 +1,200 @@
...
+    DecodeWithMemTracking,
+    TypeInfo,
+    MaxEncodedLen,
+    Debug,
+    Default,
+)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub struct ReplayWindow {
+    /// Highest nonce ever accepted for this provider (window anchor).
+    pub hwm: u64,
```
ilchu: I'd rather have this and anything around it be called more explicitly, like `highest_issued_nonce` or the like. I've never heard of highest water mark before TBH. And when it's used after a dot in some other place it made me think of a file format.

provider-node/src/command.rs
```diff
@@ -204,48 +204,62 @@ async fn start_replica_sync_coordinator(
...
-    let seed = match seed {
-        Some(s) => s.clone(),
-        None => {
-            tracing::error!("Agreement coordinator requires --keyfile for signing. Skipping.");
-            return None;
+    provider_id: &str,
+) -> Result<Arc<NonceCounter>, Box<dyn std::error::Error>> {
+    let counter = match cli.storage.storage_mode {
+        StorageMode::Inmemory => NonceCounter::in_memory(0),
+        StorageMode::Disk => {
```
ilchu: I don't see how we would need a single file with the current nonce on disk. From what I understand, it would only generate unnecessary overhead while also allowing new failure modes -- mostly due to it relying on the same atomic counter as the backing. So the state could get desynced if e.g. the counter gets updated -> the provider crashes -> the file still has the old value.

primitives/src/provider_replay_state.rs
```diff
@@ -0,0 +1,200 @@
...
+//! The LSB of `bitmap[0]` represents `hwm`, the next bit represents
+//! `hwm - 1`, and so on. Advancing the window by `d` slots shifts the
+//! bitmap left by `d` bits, dropping the oldest entries.
+
+use codec::{Decode, DecodeWithMemTracking, Encode, MaxEncodedLen};
+use core::fmt::Debug;
+use scale_info::TypeInfo;
+
+/// Width of the sliding replay window, in bits / nonce slots.
+pub const REPLAY_WINDOW_BITS: u32 = 256;
```
ilchu: Could there be an issue if the window is too small? And overall a side note on alternative design: maybe this nonce generation could be just random instead of monotonically increasing? Much less collision chance, same replay protection, less boilerplate.

ilchu (changes requested):
Some nits, some nonce-related thoughts. I still didn't go through pallet changes though, mostly provider code. So please don't merge before my approval.

provider-node/src/api.rs
danielbui12: I created follow up issue for this https://github.com/paritytech/web3-storage/issues/124

primitives/src/provider_replay_state.rs
danielbui12: @ilchu
>  And overall a side note on alternative design: maybe this nonce generation could be just random instead of monotonically increasing? Much less collision chance, same replay protection, less boilerplate.

For large-scale system, I believe monotonic is safer than random.  The "birthday paradox" problem introduces 50% collision of using random method.
FYI, I was carefully select this approach  to prevent replay among these:
- mapping(sig -> bool): Storage size grows infinitely with size of signature + 1
- Single atomic Nonce counter: Only works for simple user wallet, in this system we can have vary providers

> Could there be an issue if the window is too small? 

This is new scaffolded version. I added follow-up to https://github.com/paritytech/web3-storage/issues/124#TODO

provider-node/src/api.rs
danielbui12: Oh I see, from your recent PR

primitives/src/agreement_term.rs
danielbui12: ReplicaTerm can be in params (from user), can be in storage interface (on Provider), I guess this is unnecessary

primitives/src/agreement_term.rs
```diff
@@ -0,0 +1,51 @@
...
+use codec::{Decode, DecodeWithMemTracking, Encode, MaxEncodedLen};
+use core::fmt::Debug;
+use scale_info::TypeInfo;
+
+/// Off-chain quote signed by the provider and redeemed on-chain by the owner.
+#[derive(
+    Clone, PartialEq, Eq, Encode, Decode, DecodeWithMemTracking, TypeInfo, MaxEncodedLen, Debug,
+)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub struct AgreementTerms<AccountId, Balance, BlockNumber> {
```
mudigal, line 27: AgreementTerms does not include a bucket_id field. For primary agreements this is fine (bucket is created atomically), but for replica agreements the bucket_id is a separate extrinsic parameter (pallet/src/lib.rs - establish_replica_agreement()) not covered by the provider's signature.

A malicious owner could take a signed replica quote and redirect it to a different bucket they own. The provider ends up with a replica obligation on an unexpected bucket, potentially leading to challenge failures and slashing.

Lets add bucket_id: Option<BucketId> to AgreementTerms — None for primary, Some(id) for replica — and verify the signed value matches the extrinsic parameter.

provider-node/src/api.rs
```diff
@@ -48,6 +50,8 @@ pub fn create_router(state: Arc<ProviderState>) -> Router {
         .route("/mmr_peaks", get(get_mmr_peaks))
         .route("/mmr_subtree", get(get_mmr_subtree))
         .route("/fetch_nodes", post(fetch_nodes))
+        // Off-chain term negotiation (signed AgreementTerms for `establish_storage_agreement`)
+        .route("/negotiate", post(negotiate_terms))
```
mudigal: There can be 3 problems with some attackers - 
- Nonce exhaustion: An attacker can flood /negotiate to advance the nonce counter, causing legitimately-issued-but-not-yet-redeemed quotes to age out of the 256-bit on-chain replay window.
- CPU DoS: Each request triggers sr25519 signing.
- Price manipulation: The handler blindly signs whatever `price_per_byte` the client proposes - including 0. The on-chain extrinsic trusts the signature as consent.

Before signing, we need validate that `req.price_per_byte >= provider_settings.price_per_byte`, `req.duration` is within bounds, and `req.max_bytes` doesn't exceed capacity. Add rate limiting (e.g., `tower::limit::RateLimitLayer`).

pallet/src/lib.rs
mudigal, line 987: `complete_deregister` removes `Providers::<T>` but not `ProviderReplayStates::<T>`. This causes:
- Storage bloat: 40 bytes remain on-chain per deregistered provider
- Re-registration surprise: If the same account re-registers, old nonce windows persist and may reject fresh nonces

pallet/src/lib.rs
```diff
@@ -3255,6 +2766,39 @@ pub mod pallet {
...
+                    sp_runtime::AccountId32::new(key_bytes)
+                }
+                sp_runtime::MultiSignature::Ecdsa(_) | sp_runtime::MultiSignature::Eth(_) => {
+                    ensure!(public_key_bytes.len() == 33, Error::<T>::InvalidPublicKey);
+                    let hash = sp_io::hashing::blake2_256(public_key_bytes);
+                    sp_runtime::AccountId32::new(hash)
+                }
+            };
+
+            let hash = sp_io::hashing::blake2_256(&terms.encode());
```
mudigal: Good Practice: If any future pallet signs SCALE-encoded structs that collide byte-for-byte with AgreementTerms, cross-domain signature confusion could occur. Would it be better if `verify_terms_signature` signs `blake2_256(&terms.encode())` with some kind of a domain prefix?

provider-node/src/command.rs
```diff
@@ -204,48 +204,47 @@ async fn start_replica_sync_coordinator(
...
-
-    let seed = match seed {
-        Some(s) => s.clone(),
-        None => {
-            tracing::error!("Agreement coordinator requires --keyfile for signing. Skipping.");
-            return None;
+    provider_id: &str,
+) -> Result<Arc<NonceCounter>, Box<dyn std::error::Error>> {
+    // Start the `nonce` from 1.
+    let counter = NonceCounter::new(1);
```
mudigal: If the chain is unreachable for whatever reason, the nonce counter starts from 1 with only a warn log. 

Refuse to start `/negotiate` (set `nonce_counter` to `None`) if bootstrap fails, so the endpoint returns 503 instead of silently producing dead quotes.

provider-node/src/api.rs
```diff
@@ -729,6 +732,37 @@ async fn get_historical_roots(
...
+///
+/// TODO: Request are automatically accepted, implement advance features to let providers determine
+///
+/// Returns `503` if the node has no signing key (no `--keyfile`) or no
+async fn negotiate_terms(
+    State(state): State<Arc<ProviderState>>,
+    Json(req): Json<NegotiateRequest>,
+) -> Result<Json<SignedTerms>, Error> {
+    let keypair = state.keypair.as_ref().ok_or_else(|| {
+        Error::Internal("provider node has no signing key; /negotiate disabled".to_string())
```
mudigal: Should we be using `Error::SigningUnavailable` for the missing keypair case, instead of `Error::Internal`?

Also the documentation at the start of the function says 503 which is `Error::SigningUnavailable`.

mudigal (changes requested):
Good one overall - Some changes are needed to avoid certain attacks.

primitives/src/provider_replay_state.rs
ilchu: I'd still insist that birthday paradox would unlikely be an issue here as with a `u64` nonce we'd have to reach `u32` draws for that 50%. But upon further thinking, I guess to enforce proper TTL with random nonces we would have to keep a mapping similar to the one you described instead of the sliding window, so it's much more significant storage pressure overall. Thus I agree with monotonic nonces being used.

And some calculations re: replay window: let's say provider negotiates at rate R quotes/sec and a client holds quote N for hold time T seconds before redeeming.
Other quotes redeemed during T advance hwm; if total in-flight redemptions during T cover ≥256 fresh nonces, N is dead.

Then the safe operating regime is roughly R × T < 256, where T is the slowest legitimate client. At 1 quote/sec, T ≤ ~4 minutes. At 10/sec, T ≤ ~25 seconds. At 100/sec, T ≤ ~2.5 seconds — and that's well below realistic client decision/wallet-confirmation latency.

There's even deeper coupling between window size and `valid_until_offset`. If you set `valid_until_offset` = V blocks, the maximum number of in-flight quotes a provider can have outstanding at once is bounded by R × V × block_time. So the window is "right-sized" when:

  WINDOW_BITS ≥ R_peak × valid_until_offset × block_time

At 6s block time (and we even want to make it 2 later), `valid_until_offset` = 100, R_peak = 1/sec → need ≥ 600 slots. 256 is already too small.

So my suggestion is let's just bump it to like 1024 and forget it.

danielbui12: > So my suggestion is let's just bump it to like 1024 and forget it.

FYI, `REPLAY_WINDOW_BITS` tightly couples with `bitmap`.

```rust
   pub fn try_accept(&mut self, nonce: u64) -> Result<(), ReplayError> {
          // ....
          let distance = self.hsn - nonce;   <---- `distance` means the bit index in the `bitmap`
          if distance >= REPLAY_WINDOW_BITS as u64 {  <----- nonce older than the window can track
              return Err(ReplayError::TooOld);
          }
          let byte = (distance / 8) as usize;    <----- REPLAY_WINDOW_BITS=256 means we have 256 bits = 32 bytes, so `byte` is always in 0..32 (in-bounds for `bitmap: [u8; 32]`)
          let mask = 1u8 << (distance % 8);      <----- which bit inside that byte
          if self.bitmap[byte] & mask != 0 { <----- verify `nonce` un-used
              return Err(ReplayError::AlreadyUsed);
          }
          self.bitmap[byte] |= mask;            <----- set `nonce` to used
         // ...
      }
  ```

If we bump `REPLAY_WINDOW_BITS` to 1024, then we have to set bitmap to 1024/8 = 128 slots, which causes more storage pressure.

---

In new proposal https://github.com/paritytech/web3-storage/issues/124#TODO, I mentioned that we can have **auto submit feeless & unsigned transaction mechanism (queue + enforcing nonce must be sequence)**. 
Plus, we definitely need rate-limiting mechanism, on provider-node. let's do a basic calc:
- block_time = 6s
- 256 nonce per window
=> For 1 block, we can have, 256/6 ~ 42 nonce per provider

We have yet to measure TPS of create_bucket, create_s3_bucket, create_drive. I'm not sure the runtime can handle 42 create_bucket / create_s3_bucket / create_drive tx in a block.

Comment (ilchu) [reply primitives/src/provider_replay_state.rs]: Tbh I don't see how expanding a `u256` bitmap to 4 times that size (i.e. 128 bytes) makes a meaningful difference, considering it's bounded. E.g. even if we have 1k providers, the `ProviderReplayStates` map will still be on the order of a few MBs, and won't grow unless new providers are introduced. That should be negligible by all accounts.

I agree that rate-limiting is a good thing to have as providers are trustless and should be prevented from overwhelming the chain. But you're ignoring the `valid_until` in agreement terms (which is also currently set to `u32::MAX` automatically in the provider's API -- and that is another issue I [mentioned](https://github.com/paritytech/web3-storage/pull/105/changes#r3341562456) ). So a provider would have to honor his quotes indefinitely until claimed. And any quotes in flight that were not signed yet would have to remain valid, nonce or not, thus meaning the validity window is not per **block**, but per **block * `valid_until` period**. So if you really don't want to bump the bitmap size, then there should be much stricter limits in place for that field of agreement terms to make it reasonable for the provider.

Also from what I understand, your auto-submit + queue + enforcing nonce suggestion means that the whole bitmap and valid until field are not needed, as the requests are resolved in order. I do agree that such design is lighter and better, but then it invalidates the need to put up this sequential nonce checker in the first place. It would be much more economical to start with that design if that's the plan anyways. Or e.g. just don't implement any replay protection yet if you want to land it later.

Comment (danielbui12) [reply primitives/src/provider_replay_state.rs]: Not missing context, just want more persuasive; 
Btw, after considering. I bumped window of nonce to 1024 values; 128 bit is not a big deal, compare to other storage types

File: pallet/src/lib.rs [hunk 15]
```diff
@@ -1038,6 +1015,7 @@ pub mod pallet {
 
             T::Currency::unreserve(&who, provider.stake);
             Providers::<T>::remove(&who);
+            ProviderReplayStates::<T>::remove(&who);
```
mudigal, line 1020: I believe this is the flow for deregistering providers where ProviderReplayStates is removed. On re-registration, a fresh ReplayWindow with `hsn: 0` is inserted. This means all previously-consumed nonces become available again. If the provider re-registers with the same public key, any previously-signed quote whose `valid_until` hasn't passed can be replayed, creating new agreements and locking funds from the owner specified in the terms.

provider-node/src/command.rs
```diff
@@ -204,46 +205,76 @@ async fn start_replica_sync_coordinator(
     }
 }
 
-async fn start_agreement_coordinator(
+/// Fetch the provider's on-chain registration info once and store it for
+async fn setup_provider_info(
```
mudigal: On-chain settings (price, capacity, accepting status) are cached once at startup. If the provider updates settings on-chain, the node keeps signing terms against stale values. Related to Ilya comment on "`valid_until: u32::MAX` - terms never expire" but distinct — even with proper valid_until, stale pricing is signed.

client/src/substrate.rs
mudigal: agreement_requests_for_provider queries the AgreementRequests storage map which no longer exists in the pallet. Dynamic dispatch means it compiles but silently fails at runtime.

I am working on the https://github.com/paritytech/web3-storage/pull/139, code coverage improvement, once that PR is merged, these should get highlighted in the CI as reduction in code coverage, ideally.

pallet/src/lib.rs
```diff
@@ -4306,168 +3755,259 @@ pub mod pallet {
...
 
-            ensure!(
-                !AgreementRequests::<T>::contains_key(bucket_id, provider),
-                Error::<T>::AgreementRequestAlreadyExists
-            );
-
-            AgreementRequests::<T>::insert(bucket_id, provider, request);
+            Providers::<T>::mutate(provider, |maybe_provider| {
+                if let Some(p) = maybe_provider {
+                    p.committed_bytes = p.committed_bytes.saturating_add(terms.max_bytes);
```
mudigal: Neither `establish_storage_agreement_internal` nor `establish_replica_agreement_internal` updates the `ProviderStats::total_bytes_committed` field. It will always read 0.

pallet/src/lib.rs
```diff
@@ -4306,168 +3755,259 @@ pub mod pallet {
...
             };
 
-            ensure!(
-                !AgreementRequests::<T>::contains_key(bucket_id, provider),
-                Error::<T>::AgreementRequestAlreadyExists
-            );
-
-            AgreementRequests::<T>::insert(bucket_id, provider, request);
+            Providers::<T>::mutate(provider, |maybe_provider| {
+                if let Some(p) = maybe_provider {
```
mudigal, line 3859: The code validates via `checked_add` at line 3812 but then re-reads the provider and does `saturating_add(terms.max_bytes)` instead of assigning the already-validated new_committed. The `if let Some(p)` pattern also silently skips the update if the provider is gone.

pallet/src/lib.rs
```diff
@@ -4306,168 +3755,259 @@ pub mod pallet {
...
+            })?;
 
-            // Reserve funds
-            T::Currency::reserve(owner, total_lock)?;
+            // Validate on-chain provider's state.
+            Self::ensure_provider_active(&provider_info)?;
+            let sync_price = provider_info
+                .settings
+                .replica_sync_price
+                .ok_or(Error::<T>::ProviderNotAcceptingReplicas)?;
```
mudigal: The `sync_price` in the agreement comes from the provider's current on-chain setting, not the signed terms. The provider can change their `replica_sync_price` between signing and redemption. The owner sees one price off-chain but a different price ends up on-chain.

pallet/src/weights.rs
```diff
@@ -290,89 +249,54 @@ impl<T: frame_system::Config> WeightInfo for SubstrateWeight<T> {
...
-	fn request_agreement() -> Weight {
+	/// Storage: `StorageProvider::NextBucketId` (r:1 w:1)
+	/// Proof: `StorageProvider::NextBucketId` (`max_values`: Some(1), `max_size`: Some(8), added: 503, mode: `MaxEncodedLen`)
+	/// Storage: `StorageProvider::MemberBuckets` (r:1 w:1)
+	/// Proof: `StorageProvider::MemberBuckets` (`max_values`: None, `max_size`: Some(8050), added: 10525, mode: `MaxEncodedLen`)
+	/// Storage: `StorageProvider::Buckets` (r:0 w:1)
+	/// Proof: `StorageProvider::Buckets` (`max_values`: None, `max_size`: None, mode: `Measured`)
+	/// Storage: `StorageProvider::StorageAgreements` (r:0 w:1)
+	/// Proof: `StorageProvider::StorageAgreements` (`max_values`: None, `max_size`: Some(227), added: 2702, mode: `MaxEncodedLen`)
+	fn establish_storage_agreement() -> Weight {
```
mudigal, line 273: `establish_storage_agreement` weight says Providers has r:1 w:1, but the code does a get + mutate = 2 reads + 1 write. ProviderReplayState max_size shows 88 bytes but the actual ReplayWindow is 136 bytes (8 + 128). Both cause undercharging. Am I reading this right?

pallet/src/lib.rs
```diff
@@ -4306,168 +3755,259 @@ pub mod pallet {
...
 
-            // Check primary provider limit
-            ensure!(
-                bucket.primary_providers.len() < T::MaxPrimaryProviders::get() as usize,
-                Error::<T>::MaxPrimaryProvidersReached
-            );
+            // Primary terms must not be bound to an existing bucket — the
+            // bucket is created at redemption.
+            ensure!(terms.bucket_id.is_none(), Error::<T>::TermsBucketMismatch);
+
```
mudigal, line 3778: ensure!(terms.max_bytes > 0, ...)

provider-node/src/negotiate.rs
```diff
@@ -0,0 +1,292 @@
...
+    if info.max_capacity > 0
+        && info.committed_bytes.saturating_add(req.max_bytes) > info.max_capacity
+    {
+        return Err(Error::CapacityExceeded {
+            requested: req.max_bytes,
+            committed: info.committed_bytes,
+            max_capacity: info.max_capacity,
+        });
+    }
+
```
mudigal, line 74: check for `max_bytes > 0` is also needed isnt it?

client/src/admin.rs
```diff
@@ -116,12 +138,13 @@ impl AdminClient {
         let parsed =
             StorageProviderEventParser::from_extrinsic_events(&events, block_hash, block_number);
 
-        tracing::info!("Parsed events: {:?}", parsed);
-
         for event in parsed {
-            tracing::info!("Received event: {event:?}");
             if let StorageEvent::BucketCreated { bucket_id, .. } = event {
```
mudigal, line 142: Should we also handle the AgreementEstablished event, or document that this method is only for new-bucket creation?

provider-node/src/api.rs
```diff
@@ -48,6 +59,23 @@ pub fn create_router(state: Arc<ProviderState>) -> Router {
         .route("/mmr_peaks", get(get_mmr_peaks))
         .route("/mmr_subtree", get(get_mmr_subtree))
         .route("/fetch_nodes", post(fetch_nodes))
+        // Off-chain term negotiation (signed AgreementTerms for `establish_storage_agreement`)
+        .route(
```
mudigal, line 73: Global rate limit of 5/sec means an attacker can exhaust all 1024 replay window slots in ~3.5 minutes with anonymous requests, blocking legitimate clients from redeeming their quotes. Should we consider per-IP rate limiting or reducing the global rate?

provider-node/src/api.rs
```diff
@@ -729,6 +760,47 @@ async fn get_historical_roots(
...
+        .as_ref()
+        .and_then(|slot| slot.read().ok().map(|guard| guard.clone()))
+        .ok_or(Error::ProviderInfoUnavailable)?;
+    negotiate::validate_request(&req, &info)?;
+
+    let terms: AgreementTermsOf = AgreementTerms {
+        owner: req.owner,
+        max_bytes: req.max_bytes,
+        duration: req.duration,
+        price_per_byte: req.price_per_byte,
```
mudigal: `validate_request` ensures `req.price_per_byte >= info.price_per_byte`, but the provider signs the client's proposed price. The module doc at line 12 says otherwise. Not a security issue (client pays more by choice), but the doc is misleading.

client/src/agreement.rs
```diff
@@ -0,0 +1,146 @@
...
+    pub signature: MultiSignature,
+}
+
+/// Sign already-built terms with a provider keypair.
+///
+/// Mirror of the on-chain `verify_terms_signature`: prefix the matching
+/// term context, SCALE-encode, hash with blake2-256, then sign. The
+/// runtime accepts `MultiSignature`, so callers wrap the raw sr25519
+/// signature with `MultiSignature::Sr25519`.
+pub fn sign_terms(
```
mudigal, line 79: We have a similar function in negotiate.rs - https://github.com/paritytech/web3-storage/pull/105/changes#diff-e5127ad96fc3227f78624d209fa40318428cdb0a2e592764f65fc980aff26f82R134. Lets find a way to reuse the function and not duplicate it, please.

mudigal (changes requested):

pallet/src/lib.rs
danielbui12: @mudigal  Follow up of https://github.com/paritytech/web3-storage/pull/105#discussion_r3347239032 . The `ReplayWindow` should not have been deleted, even after provider deregistration. However, after further review, we found that the `nonce` can still be exhausted, despite having a space of 2^64, for some reasons:
- Users spam
- Users change their minds
- Users miss the deadline
- Provider Node is unhealthy
- ...
=> So I decided to remove it. & will add a extrinsic to reset it after period of time or reache to u64::MAX. 

---

Feasible solution: 
We still have RequestTimeout constant, it be applied  as max timeout for `sig. valid_until`; and the new proposed provider de-registration includes minimum announcement time,  https://github.com/paritytech/web3-storage/blob/410f4f7eac612631b1c170d344ac74afb7e8a295/pallet/src/lib.rs#L126-L128

https://github.com/paritytech/web3-storage/blob/410f4f7eac612631b1c170d344ac74afb7e8a295/pallet/src/lib.rs#L150-L155

I'll add validation to ensure that the provider deregistration period exceeds the `sig.valid_until` timeout. Once a provider decides to terminate its service, it must stop accepting new requests. By the time the `DeregisterAnnouncementPeriod` has elapsed, the last possible `RequestTimeout` will also have expired, eliminating the possibility of replaying a valid signature.

provider-node/src/command.rs
danielbui12: Agree! Not only this one, everything currently stays in cache. I also mentioned this problem in [[Provider] Investigate provider validations (role, bucket, agrement, user, ...) #88](https://github.com/paritytech/web3-storage/issues/88).

I propose we should add other workers to fully sync with the chain like Brani's draft  https://github.com/paritytech/web3-storage/pull/81.

client/src/admin.rs
danielbui12: Once a signature has been redeemed, it no longer has any value. but, the event should still be emitted for later downstream use cases (e.g., tracing and monitoring).

For the general bucket information, I don't think we need it since both events are always emitted together.

pallet/src/weights.rs
danielbui12: You can ignore this for now. we'll definitely run benchmark again once this PR is fully approved

pallet/src/lib.rs
danielbui12: btw, the provider deregistration still needs to investigate more
https://github.com/paritytech/web3-storage/issues/107#issuecomment-4646552206

provider-node/src/api.rs
danielbui12: Agree! Using global rate only increases the time required to exhaust it and even affect actual users; it doesn't eliminate the attack at all.
per-IP token bucket seems like a better mitigation since it limits replay slot consumption by any single anonymous client without impacting actual users.

provider-node/src/api.rs
danielbui12: After reviewing user's journey, using exact provider's price is solid choice.
1. Users can fill acceptable price (may above provider's price) to query all matching providers
2. But then, they can decided to choose one by provider's price.

.github/workflows/integration-tests.yml
```diff
@@ -169,37 +169,44 @@ jobs:
...
 
-      - name: Start provider (inmemory)
+      - name: Register providers on-chain
         run: |
           echo "//Alice" > /tmp/alice-key && chmod 600 /tmp/alice-key
+          echo "//Charlie" > /tmp/charlie-key && chmod 600 /tmp/charlie-key
+          cargo run --release -p storage-client --example register_provider \
+            ws://127.0.0.1:2222 http://127.0.0.1:3333 /ip4/127.0.0.1/tcp/3333 /tmp/alice-key
+          cargo run --release -p storage-client --example register_provider \
+            ws://127.0.0.1:2222 http://127.0.0.1:3334 /ip4/127.0.0.1/tcp/3334 /tmp/charlie-key
```
danielbui12, line 165: Currently, provider-node just does one-time fetch provider data from runtime. In the future, if we have  event coordinator which subscribe to runtime event & update state, this process is no longer needed.

Relevant: https://github.com/paritytech/web3-storage/pull/105#discussion_r3371102671

mudigal (approved):
## Review follow-up: deferred items + one fix pushed
  
  Catching up the open threads from my review. Most of my earlier points are
  resolved in the current revision (domain-separated signing, `bucket_id`
  binding, signed `sync_price`, `total_bytes_committed`, per-IP rate limiting,
  503 on nonce-bootstrap failure, removed stale `AgreementRequests` query,
  consolidated `sign_terms`). The remaining items are deferred and now tracked:
  
  **Tracked for follow-up**
  - **Provider-signed quotes never expire** (`valid_until = u32::MAX` in
    `negotiate_terms`) and **replay-window reset on re-registration** 
    (`complete_deregister` removes `ProviderReplayStates`; `register` reseeds
    `hsn = 0`) — these are coupled, and @danielbui12's proposed fix (bound
    `valid_until` to `RequestTimeout` + make the deregister announcement period
    exceed it) ties them to the deregistration redesign. Captured in
    #107.
  - **Provider-info cache never refreshes** — `/negotiate` validates and signs
    against settings loaded once at startup, so on-chain price/capacity/accepting
    changes aren't reflected until restart. Captured in #88
    (resolved by the chain-sync worker approach from #81).
  - **Provider-side request vetting + replay-window sizing / auto-submit queue** —
    already tracked in #124.
  - **Weights** for `establish_*_agreement` to be regenerated via the CI bench
    bot once the PR is approved (Providers r/w counts + `ReplayWindow` size).
    
## Pending in this CR 

@danielbui12 - Please fix the Code Coverage issue - Coverage needs to be at same percentages or more, surely not less.

danielbui12:
/cmd bench --runtime web3-storage-paseo --runtime storage-parachain-runtime

provider-node/src/api.rs
```diff
@@ -729,6 +781,48 @@ async fn get_historical_roots(
...
+        .ok_or(Error::ProviderInfoUnavailable)?;
+    negotiate::validate_request(&req, &info)?;
+
+    let terms: AgreementTermsOf = AgreementTerms {
+        owner: req.owner,
+        max_bytes: req.max_bytes,
+        duration: req.duration,
+        price_per_byte: info.price_per_byte,
+        // TODO: current_block + StorageProvider::RequestTimeout
+        valid_until: u32::MAX,
```
ilchu, line 816: I feel this one is sufficiently small to do in this PR, and would not leave it as a TODO. Or at least have a tracking issue for this if you really want to merge ASAP.

ilchu (approved):
This PR has taken a nice shape, that was a lot of work, so good job! AFAICT the pallet code is fine too, no blockers found. Only one comment that I'd like to be dealt with, otherwise LGTM.

provider-node/src/api.rs
danielbui12: I planned to complete a event coordinator job, that hold runtime state; after that negotiate function can get block number from it, instead of repeatedly initialize subxt::OnlineClient everywhere

PR #106, opened 2026-05-29
fix(bot): install missing `frame-omni-bencher` in CI for bench

ilchu:
As the cmd bot is failing benchmark due to `frame-omni-bencher` being absent, download just that tagged binary first in CI. Bulletin has a much larger action for bringing in all the polkadot binaries, but here we don't need that much yet.

### DIFF
--- a/.github/workflows/cmd-run.yml
+++ b/.github/workflows/cmd-run.yml
@@ -138,6 +138,17 @@ jobs:
           repository: ${{ env.REPO }}
           ref: ${{ env.PR_BRANCH }}
 
+      - name: Install frame-omni-bencher for bench
+        if: startsWith(github.event.inputs.cmd, 'bench')
+        shell: bash
+        run: |
+          set -euxo pipefail
+          source .github/env
+          curl -sSfL -o /usr/local/bin/frame-omni-bencher \
+            "https://github.com/paritytech/polkadot-sdk/releases/download/${POLKADOT_SDK_VERSION}/frame-omni-bencher"
+          chmod +x /usr/local/bin/frame-omni-bencher
+          frame-omni-bencher --version
+
       - name: Run cmd
         id: cmd
         env:


.github/workflows/cmd-run.yml
```diff
@@ -138,6 +138,17 @@ jobs:
...
+        if: startsWith(github.event.inputs.cmd, 'bench')
+        shell: bash
+        run: |
+          set -euxo pipefail
+          source .github/env
+          curl -sSfL -o /usr/local/bin/frame-omni-bencher \
+            "https://github.com/paritytech/polkadot-sdk/releases/download/${POLKADOT_SDK_VERSION}/frame-omni-bencher"
+          chmod +x /usr/local/bin/frame-omni-bencher
+          frame-omni-bencher --version
+
```
danielbui12, line 151: I believe we can cache this binary because of not frequent change.
For example:
```yaml
- name: Cache frame-omni-bencher
  if: startsWith(github.event.inputs.cmd, 'bench')
  id: cache-frame-omni-bencher
  uses: actions/cache@v4
  with:
    path: /usr/local/bin/frame-omni-bencher
    key: frame-omni-bencher-${{ runner.os }}-${{ env.POLKADOT_SDK_VERSION }}

- name: Install frame-omni-bencher for bench
  if: startsWith(github.event.inputs.cmd, 'bench') && steps.cache-frame-omni-bencher.outputs.cache-hit != 'true'
  shell: bash
  run: |
    set -euxo pipefail
    curl -sSfL -o /usr/local/bin/frame-omni-bencher \
      "https://github.com/paritytech/polkadot-sdk/releases/download/${POLKADOT_SDK_VERSION}/frame-omni-bencher"
    chmod +x /usr/local/bin/frame-omni-bencher

- name: Verify frame-omni-bencher
  if: startsWith(github.event.inputs.cmd, 'bench')
  shell: bash
  run: |
    set -euxo pipefail
    frame-omni-bencher --version
```

danielbui12: Or we can even create a re-useable workflow file to cache all polkadot binaries then use them across all workflow files https://github.com/paritytech/web3-storage/blob/dev/.github/workflows/integration-tests.yml#L63-L94, like the one in bulletin chain https://github.com/paritytech/polkadot-bulletin-chain/blob/main/.github/actions/use-polkadot-binaries/action.yml

ilchu: Yeah, caching would be good. And it can be another just recipe. I'll follow up.

PR #110, opened 2026-06-01
Hard-fail on provider CID mismatch instead of warn

ilchu:
The `create_drive` function previously logged a `tracing::warn!` and continued when the provider's returned data_root disagreed with the locally-computed CID, defeating content-addressing's only integrity guarantee.

Now returns a new `FsClientError::CidMismatch` and refuses to cache or proceed. Adds unit coverage for both the matching and mismatching path via a small `verify_cid` helper.


storage-interfaces/file-system/client/src/lib.rs
```diff
@@ -807,6 +809,16 @@ impl FileSystemClient {
         Ok(data)
     }
 
+    /// Compare a locally-computed CID against the CID a provider returned for
+    /// the same bytes. Returns `CidMismatch` on disagreement so callers can
+    /// refuse to trust the provider's response.
+    fn verify_cid(expected: Cid, got: Cid) -> Result<()> {
```
bkontur: @ilchu for this kind of check, we are using `ensure_**` if you check PolkadotSDK, so I would vote for `ensure_cid_matches` or better name :)

```suggestion
    fn ensure_cid_matches(expected: Cid, got: Cid) -> Result<()> {
```

bkontur (approved):
@ilchu please check just: https://github.com/paritytech/web3-storage/pull/110/files#r3341175555

Issue #107, opened 2026-06-01
Investigate optimal provider deregistration


ilchu:
Currently we made provider deregistration 2-step (#80), but that may still be suboptimal. We need to think about ease of use locally as well, as e.g. #96 showed that the state remains polluted between test runs due to deregistration period not having passed. In that case it was circumvented with a sudo fast-forward, but there might be something more optimal overall. 

<details>
<summary>Details</summary>

> Fair point for the runtime, and good to know it's covered. Here though the problem was for local reruns on the same global setup: the account used for acknowledging the UI state didn't completely deregister, so it never saw the right screen. But I guess instead of skipping that particular test if the Ferdie account is already there like I did initially I'll try sudo removing the thing to make tests more reproducible. Not that it's a CI problem, but still.

@ilchu For decentralization, I would not use sudo on specific logic.
Originally, the provider deregistration flow is just check if provider's commit_bytes is 0; to be more specific we should check provider is not in charge any buckets. This way provider can instantly leave; but to do that, we have to iterate the whole buckets storage to examine (not cost effective).

My fixes introduce provider deregistration in 2 steps may not ideal as well. However, we can fallback to original with additional storage mapping to trace provider's buckets easier, or when look deeper there are better solutions for this.

Let's open a issue for this, we'll investigate more.

_Originally posted by @danielbui12 in https://github.com/paritytech/web3-storage/issues/96#issuecomment-4569305186_

</details>

danielbui12:
More context: 

1. If a provider was primary provider for a bucket, and provider would wanted to leave instantly?
- When can provider completely leave?
- What happen with the bucket? If we allowed the provide to leave
  - Are there any slash mechanisms in this case?
  - System alternatively finds a new primary provider and bucket should be replicated to the new provider before current provider can leave

2. The pallet constant T::DeregisterAnnouncementPeriod is not guaranteed to be the longest period among all buckets.

danielbui12:
I find out that `committed_bytes` already ensure provider has no bucket obligation left.

Here is brief about its lifecycle:
It's a running counter on ProviderInfo that mirrors the provider's total contracted bytes across all live agreements. It moves in lockstep with StorageAgreements rows — every row added bumps it up, every row removed brings it down.

- Goes up (always paired with a StorageAgreements::insert):
  - Agreement accepted — +max_bytes, on both accept paths (pallet/lib.rs:1339, pallet/lib.rs:1835).
[...]

Issue #44, opened 2026-03-25
Production readiness: Web3 principles compliance and Polkadot SDK standards


mudigal:
# Production Readiness: Web3 Principles Compliance & Polkadot SDK Standards

This issue tracks all architectural changes required to align Web3 Storage with Web3 principles ("less trust, more truth") and Polkadot SDK production standards. Each item includes the problem, impact, and proposed solution.

---

## Priority Legend

| Priority | Meaning |
|----------|---------|
| 🔴 P0 | Blocks production deployment — must fix |
| 🟠 P1 | Serious gap — fix before testnet launch |
| 🟡 P2 | Important improvement — fix before mainnet |
| 🟢 P3 | Good practice — fix when convenient |

---

### A1. 🔴 Client-Side Encryption (User Sovereignty)

**Problem**: Data is stored in plaintext on the provider's disk. The provider can read, copy, or leak every byte. Users have zero privacy from the entity storing their data.

**Solution**: Implement mandatory client-side encryption before upload.

**Design**:
- Add an `encryption` module to `client/src/` and the Console UI `StorageClient`
- Use **XChaCha20-Poly1305** (AEAD) for symmetric encryption of each chunk before upload
- Key derivation: **x25519** key exchange between client keypair and a per-bucket ephemeral key, or a user-supplied passphrase via **Argon2id** KDF
- Encryption is applied **before** chunking and hashing — the provider only ever sees ciphertext
- Decryption key is never sent to the provider — stored client-side (optionally backed up encrypted on-chain via a small metadata field)
- The MMR commitment is over encrypted chunks, so challenge/response works unchanged

**Files to change**:
- `client/src/encryption.rs` (new) — `encrypt_chunk(key, plaintext) → ciphertext`, `decrypt_chunk(key, ciphertext) → plaintext`
- `client/src/storage_user.rs` — wrap upload/download with encrypt/decrypt
- `user-interfaces/console-ui/src/lib/storage.ts` — add WebCrypto-based encryption before `putObject`
- `primitives/src/lib.rs` — add `EncryptionMetadata` type (algorithm, nonce, key hint)

**Scope**: ~500 LOC Rust, ~200 LOC TypeScript

---

### A2. 🔴 Download Integrity Verification (Trustlessness)

**Problem**: When the client downloads data, there is no cryptographic verification that the bytes match the committed hash. A malicious provider can serve garbage.

**Solution**: Verify content hash on every download and optionally request Merkle proofs.

**Design**:
[...]

PR #125, opened 2026-06-03
Challenge flow overhaul

### LINKED ISSUE #225: Challenge flow overhaul
_Tracking issue from the cross-repo roadmap plan (2026-06-23)._

Track the challenge flow overhaul (PR #125). Core protocol for Web3 mainnet.

**Covers PR(s):** #125
**Theme:** Challenge Protocol & Economics · **Category:** Feature
**Priority:** P1 · **Effort:** L (~8 wk) · **Target:** Web3 Mainnet (Dec '26)

> [!NOTE]
> Relates to #157 (cost escalation).

ilchu:
The provider was where trust was supposed to land but the chain couldn't actually punish a provider that lied — `respond_to_challenge` slashed only on timeout, signed commitments could be replayed forever, and several pallet placeholders made the math wrong. This PR closes all of those.

Seven independently-reviewable commits, starting from a baseline test commit so the security-critical changes land into a covered area (the pallet had zero challenge tests before this).

  ## Pallet — security

  **Slash on demonstrably-false response, not just on timeout.** `respond_to_challenge` now adjudicates each variant into `Ok(())` or `Err(SlashReason)`. Invalid `Proof` (chunk-Merkle or MMR mismatch), forged `Deleted` admin signature, and bogus `Superseded` claim all slash immediately. Parameter-shape errors (stale nonce, non-admin signer) still bubble as `DispatchError` — those are caller mistakes, not lies. New `SlashReason` enum carried in the `ChallengeSlashed` event.

  **Replay protection on `CommitmentPayload`.** New `nonce: u64` field; `CURRENT_VERSION` bumped to `0x02` so V1-shaped signatures can't verify against the V2 layout. New `T::MaxNonceAge` runtime constant (24h on live runtimes), recency-checked on every signature path — `checkpoint`, `extend_checkpoint`, `challenge_offchain`, `Deleted` response arm. `BucketSnapshot` carries `commitment_nonce` so late-arriving `extend_checkpoint` signatures verify against the same payload the original signers signed.

  **Three placeholder fixes.** `challenge_offchain` now takes `leaf_count: u64` from the challenger rather than reconstructing with hardcoded `0` (which only worked if the provider also signed with 0). `Replica.last_sync` is now `Option<ReplicaSyncRecord {
  mmr_root, start_seq, leaf_count, block }>` populated from the bucket snapshot at `confirm_replica_sync` time; `challenge_replica` uses the stored `start_seq` instead of `0u64`. Challenge deposit is now `T::ChallengeDeposit` (1 token live, was `100u32` ≈ 1e-10 of a token — effectively free spam).

  ## Pallet — observability

  **`ChallengerStats` on-chain aggregate.** New storage map `ChallengerStats<AccountId, ChallengerStatRecord>` with `ValueQuery`. Bumped at four sites: `create_challenge` increments `total_challenges`; defended path increments `failed_challenges`; `slash_provider_for_failed_challenge` (used by both timeout and invalid-response paths) increments `successful_challenges` and adds the 10%-of-slashed-stake reward to `total_earnings`.

  ## Provider node

  **Wire `ChallengeResponder` into startup.** The module was defined and exported but never instantiated in `command.rs` — entire path was dead code in production. New `--enable-challenge-responder` CLI flag alongside the other coordinators.

  **Real `poll_challenges`.** Was `Ok(vec![])`. Now resolves the provider's 32-byte account, iterates `StorageProvider::Challenges` via subxt, manually decodes the SCALE-encoded `Vec<Challenge>`, and keeps only entries matching our account (preserving the original index for `ChallengeId` reconstruction). Cost is bounded by `ChallengeTimeout` — past-deadline entries are reaped in `on_finalize`.

  **API surface.** `/commit`, `/commitment`, `/checkpoint-signature`, and `/delete` now take a caller-supplied `nonce` (the block the caller intends to submit at) and echo it in the response. Signed payloads use the real post-mutation `leaf_count` (was 0).

  ## Client SDK

  `get_total_challenge_earnings` and `get_challenge_stats` are real — both read `ChallengerStats` from chain. `commit` / `get_commitment` / `get_checkpoint_signature` take an explicit `nonce: u64` so callers control what block the signature is bound to.

  ## JS demos

  `examples/papi/{full-flow,bucket-with-storage,checkpoint-rewards,sc-flow,sc-coverage}.js` snapshot `System.Number` and thread the value through `uploadChunk`, `fetchCheckpointSignature`, `challengeOffchain`, `submitClientCheckpoint`. `upload.leafCount` and `upload.nonce` from the `/commit` response are passed to `challenge_offchain`.

  ## Tests

  22 new challenge tests in the pallet (up from 0) covering baseline happy/timeout/no-snapshot cases, replay-rejection for `challenge_offchain` and `checkpoint`, immediate-slash for all three invalid response variants, `ChallengerStats` updates on each transition, and `challenge_replica` using the stored `start_seq`. New SCALE-decoder round-trip unit test in `challenge_responder.rs`. Provider-node integration tests updated for the new nonce / leaf_count surface. Client integration tests no longer hardcode "earnings = 0" — they assert the real `total >= successful + failed` invariant.

  ## Migration

  Breaking: `CommitmentPayload` v1 signatures (no nonce) no longer verify on this runtime. Acceptable for a pre-launch parachain — dev-chain state is expected to break. Extrinsic signatures have new args (`nonce`, `leaf_count`); JS clients, Rust SDK callers, and benchmarks are all updated.

---

_Closes #225 — cross-repo roadmap tracking._


bkontur:
@ilchu do you want to get this in or after w3s?

ilchu:
> @ilchu do you want to get this in or after w3s?

No rush, I just wanted this up to date. This is mostly a correctness issue, so not needed for a POC demo per se.

pallet/src/lib.rs
```diff
@@ -199,12 +288,68 @@ pub mod pallet {
...
+    /// reused for a given deadline, so a `ChallengeId { deadline, index }`
+    /// stays valid even when sibling challenges sharing the same deadline are
+    /// resolved (the old `Vec`-backed layout shifted indices on removal,
+    /// making siblings unaddressable).
     #[pallet::storage]
-    #[pallet::unbounded]
     #[pallet::getter(fn challenges)]
-    pub type Challenges<T: Config> =
-        StorageMap<_, Blake2_128Concat, BlockNumberFor<T>, Vec<Challenge<T>>>;
+    pub type Challenges<T: Config> = StorageDoubleMap<
```
bkontur, line 298: @ilchu hmm, why it did not hit check-migration for PreviewNet? no challanges there?

provider-node/src/subxt_client.rs
```diff
@@ -805,17 +807,79 @@ impl ReplicaSyncChainClient for SubxtChainClient {
...
+            // DoubleMap key layout:
+            //   16 (pallet hash) + 16 (storage hash)
+            //   + 16 (blake2_128 of deadline) + 4 (BlockNumber u32 LE)
+            //   + 8 (twox64 of index)         + 2 (u16 index LE)
+            //   = 62 bytes total.
+            let key_bytes = &kv.key_bytes;
+            if key_bytes.len() < 62 {
+                continue;
+            }
+            let deadline = u32::from_le_bytes(key_bytes[48..52].try_into().unwrap_or([0; 4]));
```
bkontur, line 857: hmm, this direct storage key parsing is kind of weak and fragile, for example what if the pallet changes the format.

I would do this call over RuntimeApi - expose versioned runtime api, if the format or struct changes, we just bump runtime api and the provider will fail here - please fix or create follow-up issue, do we use this storage item in provider anywhere else?

bkontur: @ilchu looks what I've found [here](https://github.com/paritytech/web3-storage/issues/44): B6. 🟠 Add Custom Runtime APIs here  (didnt check other suggestions)

pallet/src/lib.rs
ilchu: I guess that makes sense, don't see any other immediate explanation.

pallet/src/lib.rs
```diff
@@ -131,29 +131,25 @@ pub mod pallet {
...
             // more blocks), so an old quote cannot be replayed against the new
             // incarnation.
+            // At the same time, the deregistration announcement window must be at least
+            // as long as the challenge response timeout, so any challenge created
+            // up to the announcement block matures (and the provider stays
+            // slashable) before the provider can complete deregistration.
             assert!(
-                T::RequestTimeout::get() < T::DeregisterAnnouncementPeriod::get(),
-                "RequestTimeout must be less than DeregisterAnnouncementPeriod to close the re-register replay window"
+                T::RequestTimeout::get() < T::DeregisterAnnouncementPeriod::get()
```
bkontur, line 136: I marked this in @danielbui12's PR, we should do this checks also in try_state - follow up https://github.com/paritytech/web3-storage/issues/230

provider-node/src/subxt_client.rs
```diff
@@ -805,17 +807,79 @@ impl ReplicaSyncChainClient for SubxtChainClient {
...
+            detected.push(DetectedChallenge {
+                bucket_id: challenge.bucket_id,
+                deadline,
+                index,
+                mmr_root: challenge.mmr_root,
+                start_seq: challenge.start_seq,
+                leaf_index: challenge.leaf_index,
+                chunk_index: challenge.chunk_index,
+                challenger: sp_core::crypto::AccountId32::from(challenge.challenger).to_ss58check(),
+                created_at_block: deadline,
```
bkontur: @ilchu  looks like this `created_at_block` is not used anywhere, do we need it? Can we remove that from `DetectedChallenge`?
```suggestion
```

pallet/src/lib.rs
ilchu: Yeah, I've seen it, I'll assign it to myself as a followup.

provider-node/src/subxt_client.rs
ilchu: Yeah, I guess it's not needed when we have deadline field. Also this assignment itself was incorrect, thanks for pointing out.

provider-node/src/subxt_client.rs
ilchu: Sure, there's also quite a bit of such ad hoc parsing in the client, so I believe it's better to fix it in one sweep. Created #237 for this.

provider-node/src/subxt_client.rs
bkontur: > Yeah, I guess it's not needed when we have deadline field. Also this assignment itself was incorrect, thanks for pointing out.

ok, nice, let's remove created_at_block

pallet/src/lib.rs
```diff
@@ -68,19 +75,60 @@ pub mod pallet {
...
+            //   reads  (10): `drain_prefix` next entry (1); `decrement_pending`
+            //     reads `PendingChallenges` + `PendingChallengesByBucket` (2);
+            //     `slash_provider_for_failed_challenge` reads `Providers` (1),
+            //     the slash + unreserve + two `resolve_creating` currency ops
+            //     (4), and `ChallengerStats` (1), plus 1 slack.
+            //   writes (10): `drain_prefix` removes the entry (1);
+            //     `decrement_pending` writes both pending counters (2); the
+            //     currency slash/unreserve/two reward settlements (4);
+            //     `Providers::insert` (1); `ChallengerStats` (1), plus 1 slack.
+            let per_challenge = T::DbWeight::get().reads_writes(10, 10);
```
bkontur: @ilchu there is lots of assumptions in this doc, this will very quickly get out of the sync, also using hard-coded read/writes could be tricky and not that safe, the better solution would be to create dedicated benchmark and use benchmarked exact weight like: `T::WeightInfo::on_initialize_slash_challanges(count)` or something like that, we should provider exact measured benchmarked value per runtime

primitives/src/lib.rs
```diff
@@ -147,6 +173,69 @@ pub enum RemovalReason {
...
+    /// Total challenges the challenger has ever opened.
+    pub total_challenges: u32,
+    /// Challenges where the provider was slashed (either invalid response
+    /// or timeout). These pay the challenger.
+    pub successful_challenges: u32,
+    /// Challenges where the provider successfully defended.
+    pub failed_challenges: u32,
+    /// Cumulative rewards earned from successful challenges (provider stake
+    /// slash share). Refunded deposits are NOT counted here.
+    pub total_earnings: Balance,
```
bkontur: @ilchu description says:
> increments successful_challenges and adds the 10%-of-slashed-stake reward to total_earnings.

but it looks like the `total_earnings` is never modified anywhere or accumulated, can you please check this?

ilchu: It was a remnant of the old way where we assumed challenges could be rewarded, so both this field and all comments should have been gone. Fixed now.

danielbui12:
@ilchu Could you also follow up my latest comment **"Double check last Challenge expiration (ChallengeTimeout) must smaller than DeregisterAnnouncementPeriod"** on this issue https://github.com/paritytech/web3-storage/issues/107#issuecomment-4725782964 then close it

pallet/src/lib.rs
```diff
@@ -1819,19 +2000,25 @@ pub mod pallet {
         /// Submit a new checkpoint with provider signatures.
         #[pallet::call_index(30)]
         #[pallet::weight(T::WeightInfo::checkpoint())]
+        #[allow(clippy::too_many_arguments)]
```
bkontur, line 2004: @ilchu (follow up to remove `clippy::too_many_arguments`) please check if we could wrap some params logically together, for example I see:
```
            mmr_root: H256,
            start_seq: u64,
            leaf_count: u64,
```
used also for another extrinsic `provider_checkpoint` and also similar combination is used for another ones:
```
mmr_root: H256,
start_seq: u64
leaf_index: u64
```
not sure if leaf_count is the same as leaf_index (probably no), but we could check for possible deduplications, e.g. lots of pallet Events also has those triplets and so on

pallet/src/benchmarking.rs
```diff
@@ -1171,5 +1195,47 @@ mod benchmarks {
...
 
+    /// `on_finalize` slash sweep: drains and slashes every challenge expiring
+    /// at a deadline. Linear in the challenge count `c` (bounded at runtime by
+    /// `MaxChallengesPerDeadline`); each entry is drained, its pending counters
+    /// decremented, and its provider slashed. The cost is charged up front in
+    /// `on_initialize` via `WeightInfo::on_initialize_slash_challenges(c)`. The
+    /// work is independent per challenge, so the linear fit extrapolates to the
+    /// runtime cap.
+    #[benchmark]
+    fn on_initialize_slash_challenges(c: Linear<0, 100>) {
```
bkontur: @ilchu Shoudn't we align this hard-coded 100 with runtime configuration which is set to [1000](https://github.com/paritytech/web3-storage/pull/125/changes#diff-adb706e74e9931b2d9af4b55c44b94943353a52b0dc291045a25b13fc138c686R49)? 

```suggestion
    fn on_initialize_slash_challenges(c: Linear<0, 1000>) {
```

there is possibility to read the runtime config with something like this:
```
	u: Linear<0, { T::MaxChallengesPerDeadline::get() }
```

Or 1000 how was this stated? Can we lower to 100?

bkontur (approved):
ok modulo please check this [comment](https://github.com/paritytech/web3-storage/pull/125#discussion_r3490975823)

ilchu:
> @ilchu Could you also follow up my latest comment **"Double check last Challenge expiration (ChallengeTimeout) must smaller than DeregisterAnnouncementPeriod"** on this issue [#107 (comment)](https://github.com/paritytech/web3-storage/issues/107#issuecomment-4725782964) then close it

Fixed in 707318e. Made the inequality strict and added some additional checks, so should be good.

pallet/src/lib.rs
ilchu: This one turns out to be a bigger sweep than I thought, so I'll do it separately. Tracked in #250.

pallet/src/benchmarking.rs
ilchu: I think 1000 itself is an arbitrary number, but it is our current max challenge cap. So I've adjust it to be 1000 (via the getter) here too.

danielbui12:
@ilchu Since you updated the checkpoint extrinsic, don't forget to check the `client/src/substrate.rs` as well. I'll fix it in my PR this time

PR #140, opened 2026-06-05
try-runtime job for PreviewNet

### LINKED ISSUE #103: [CI] Add try-runtime CI job, runtime matrix, check for PreviewNet
We are live in PreviewNet, so all the changes should check if we need migration or not

rosarp:
Closes #103


.github/env
```diff
@@ -1,6 +1,7 @@
 RUST_STABLE_VERSION=1.93.0
 POLKADOT_SDK_VERSION=polkadot-stable2603
 ZOMBIENET_VERSION=v0.4.11
+TRY_RUNTIME_VERSION=v0.10.1
```
ilchu, line 4: Would we also want to pin the SHA256 like in bulletin?

justfile
```diff
@@ -9,6 +9,8 @@
 polkadot_version := "polkadot-stable2603"
 # Zombienet version
 zombienet_version := "v0.4.11"
+# try-runtime CLI version (for runtime migration checks)
+try_runtime_version := "v0.10.1"
```
ilchu: Orthogonal to this PR, so not a blocker: how about having `.github/env` as the single source of truth like in bulletin, i.e.:
```just
# Shell-exported vars still override file values.
set dotenv-load := true
set dotenv-filename := ".github/env"
```

ilchu (approved):
Nothing blocking, but I did find a few followups. Feel free to add them in this PR or I'd appreciate if you could create issues.

The biggest one that wasn't your code related: across the 3 main pallets (storage provider, s3 registry and drive registry) the latter doesn't have the `try-runtime` feature at all. We'd definitely need that for completeness of this CI check, but whether you want to do it separately or on top of this is up to you.

scripts/runtimes-matrix.json
```diff
@@ -25,14 +24,13 @@
     "path": "runtime",
     "build_command": "build-runtime",
     "zombienet_config": "zombienet.toml",
-    "chain_spec_script": "scripts/build-chain-spec.sh",
     "uris": [
-      "TODO: update WSS"
+      "wss://previewnet.substrate.dev/web3-storage"
```
bkontur:
```suggestion
      "TODO: update WSS"
```

scripts/runtimes-matrix.json
```diff
@@ -5,14 +5,13 @@
     "path": "runtimes/web3-storage-paseo",
     "build_command": "build-paseo-runtime",
     "zombienet_config": "zombienet/storage-paseo-local.toml",
-    "chain_spec_script": "scripts/build-paseo-chain-spec.sh",
     "uris": [
       "TODO: update Paseo WSS"
```
bkontur:
```suggestion
      "wss://previewnet.substrate.dev/web3-storage"
```

bkontur (changes requested):
@rosarp or @ilchu please also check https://github.com/paritytech/web3-storage/actions/runs/27340155044/job/80774562618?pr=140

ilchu (approved):

bkontur:
actually, it does not run check migrations job: https://github.com/paritytech/web3-storage/actions/runs/27536521335 :)

bkontur (approved):

bkontur:
@rosarp @ilchu nice, thank you, now passes: https://github.com/paritytech/web3-storage/actions/runs/27546257023/job/81421099082?pr=140 (making it as required)

mudigal:
I would appreciate if Code Coverage was maintained. That is the reason it was introduced in basic check.

PR #141, opened 2026-06-05
ci: fold ui-e2e workflow into integration-tests

### LINKED ISSUE #137: [CI] Make ui-e2e.yml as a part of integration-tests.yml
On the top of https://github.com/paritytech/web3-storage/pull/129

```diff
  integration-tests-complete:
    name: Integration Tests
-    needs: [build, integration-tests, e2e-integration-tests, sc-integration-tests, coverage-gate]
+    needs: [build, integration-tests, e2e-integration-tests, sc-integration-tests, ui-integration-tests, coverage-gate]
```

ilchu:
Closes #137.

Folds the standalone `ui-e2e.yml` Playwright workflow into `integration-tests.yml` as a `ui-integration-tests` job, gated by the existing `integration-tests-complete` aggregate — so the required **Integration Tests** check now covers UI e2e too, and there's no second workflow / second required check to maintain.

### Changes
- **New `ui-integration-tests` job** in `integration-tests.yml`: drive-ui / console-ui / provider Playwright suites against a paseo chain + inmemory provider, modelled on the existing `integration-tests` job.
- Added `ui-integration-tests` to `integration-tests-complete`'s `needs`.
- **Deleted `ui-e2e.yml`** — its duplicate `set-image`, the `changes` paths-filter, the `ui-e2e-gate`, the raw `zombienet spawn`, and the hand-rolled provider health loop all go away.
- README CI-gates section updated to point at the new job.

### Faster
The old workflow ran a full `just build-paseo-runtime && just build-provider` on every run. The new job consumes the shared `build` artifact (provider binary + runtime wasm) and the warm `integration-tests-build` cache like its siblings, so the chain-spec script's `cargo build` is a near-no-op. It runs in parallel with the other artifact-consuming test jobs (`needs: [set-image, build]`), so it doesn't extend the workflow's critical path. UIs are served by Vite dev servers, so there's no UI build step.

### Behaviour change
UI e2e is no longer path-gated — it now runs on every PR, consistent with the rest of the integration suite (which has never had a path filter). Doc-only PRs already run the full build + integration matrix, so this is consistent, not new cost.


.github/workflows/integration-tests.yml
```diff
@@ -539,6 +539,177 @@ jobs:
...
+      - name: Install workspace deps
+        working-directory: user-interfaces
+        run: pnpm install --frozen-lockfile
+
+      - name: Install Playwright browsers
+        working-directory: user-interfaces
+        run: pnpm --filter @web3-storage/drive-ui exec playwright install --with-deps chromium
+
+      - name: Start chain
+        uses: ./.github/actions/start-zombienet
```
bkontur: @ilchu please tell you assistant to switch to the `start-e2e-chain` - it will clean/remove zombienet cache and logs stuff around

bkontur (approved):
lgtm modulo https://github.com/paritytech/web3-storage/pull/141/changes#r3386753002

.github/workflows/integration-tests.yml
```diff
@@ -505,6 +505,176 @@ jobs:
...
+        run: cargo install just --locked || true
+
+      - name: Cache Polkadot SDK binaries
+        uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5
+        id: sdk-cache
+        with:
+          path: |
+            .bin/polkadot
+            .bin/polkadot-execute-worker
+            .bin/polkadot-prepare-worker
```
bkontur, line 552: for `start-e2e-chain` we don't need this, but this is super/ultra nit :)

```suggestion
```

PR #142, opened 2026-06-08
Add user-interfaces and github actions to dependabot

RafalMirowski1:
Adds dependabot support for user-interfaces, examples and github actions, mostly for the future as I think by default it will only open prs against main (though we can set it to cover dev too if we want)


bkontur (approved):

bkontur (changes requested):
@RafalMirowski1 Cisco's new github actions audit checker reports: https://github.com/paritytech/web3-storage/actions/runs/27339363724/job/80771856669?pr=142

bkontur (approved):

Issue #144, opened 2026-06-08
Improve code coverage across pallets, provider-node, and CI


mudigal:
## Objective

Improve test coverage for pallets (~73% → 80%+) and provider-node (~73% → 80%+) by:
1. Extracting chain interaction traits for mock-based testing
2. Adding comprehensive integration tests
3. Hardening CI coverage measurement

### PR 1: `refactor(provider-node): extract chain traits for testability and harden CI`
**Foundational refactoring — no tests.**
- Extract `AgreementChainClient`, `ChallengeChainClient`, `CheckpointChainClient`, `ReplicaSyncChainClient` traits from concrete Subxt implementations
- Move Subxt implementations to separate `_subxt.rs` files (excluded from coverage)
- Add unified `Error` type with proper HTTP status mapping
- Update CI: measure all 3 pallets combined, glob-based test discovery, validation guards
- Fix S3 CI nonce contention (use `//Bob` instead of `//Alice`)
- Fix CI flakiness in integration tests

### PR 2: `test(pallet): add 47 tests for challenges, checkpoints, replicas, events, and runtime API`
**Independent — no dependency on PR 1.**
- Challenge cost-split boundary tests (5 threshold levels)
- Checkpoint scheduling: leader election, grace fallback, missed checkpoint, pool funding
- Replica sync lifecycle: request, confirm, reject, data root mismatch
- Runtime API: provider info, bucket info, agreement info, can_accept_bytes, find_matching_providers
- Event emission verification for all state-changing extrinsics
- Warning comment on `run_to_block` about missing pallet hooks

### PR 3: `test(provider-node): add HTTP and storage integration tests`
**Stacked on PR 1.**
- Full HTTP request/response tests for all API endpoints
- Authentication middleware tests (valid/invalid/expired tokens, bucket membership)
- Disk storage backend tests (write, read, list, delete, edge cases)
- File system API tests (directory operations, file upload/download)
- S3 API tests (bucket operations, object lifecycle)

### PR 4: `test: add coordinator mock tests and file-system pallet tests`
**Stacked on PR 3.**
- Agreement coordinator: poll-and-accept, error handling, auto-accept toggle
- Challenge responder: successful response, proof generation failure, submission failure, pause/resume
- Checkpoint coordinator: duty detection, force checkpoint, pause/resume
- Replica sync coordinator: sync lifecycle, endpoint resolution, balance checks
- File-system drive-registry pallet: success-path tests for create/delete/share/unshare

[...]

PR #139, opened 2026-06-05
Code Coverage Improvement - avoiding dead codes, improving test scenarios and cases 

mudigal:
## Summary

Comprehensive test coverage improvement across the entire codebase, raising pallet coverage from 62% to 80% and provider-node coverage from 26% to 78%.

### Pallet tests
- **47 new tests** for `pallet-storage-provider`: challenge costs, checkpoints, replicas, events, and runtime API
- **8 new success-path tests** for `pallet-drive-registry`: drive CRUD, sharing/unsharing, admin delegation (exercises previously-uncovered `_internal` functions)

### Provider-node tests
- **50+ new integration tests** across API endpoints, file system, S3, error paths, edge cases, and storage utilities
- **6 new challenge responder tests** with realistic storage data setup and proof generation
- Extracted chain traits from background coordinators for testability
- Moved coordinator mock tests to integration tests for CI coverage visibility

### CI improvements
- Measure all 3 pallets together (`pallet-storage-provider`, `pallet-drive-registry`, `pallet-s3-registry`) in a single coverage run
- Dynamically discover integration test targets for provider-node coverage
- Fix Zombienet smoke test flakiness: use `//Bob` for S3 demo to avoid nonce contention with provider node running as `//Alice`
- Reduce SC test flakiness by switching to dev-mode and simplifying CI matrix

## Test plan

- [x] `cargo test --workspace` — all tests pass
- [x] `cargo clippy --all-targets --all-features --workspace -- -D warnings` — clean
- [x] CI coverage gate passes (62% → 80% pallet, 26% → 78% provider)

PR #145, opened 2026-06-08
refactor(provider-node): extract chain traits for testability and harden CI

mudigal:
## Summary

https://github.com/paritytech/web3-storage/issues/144

Split from #139 (PR 1 of 4).

- **Trait extraction**: Extract chain interaction traits from 4 coordinators (agreement, challenge, checkpoint, replica-sync) so each can be tested with mock chain clients. Move Subxt implementations to dedicated `_subxt.rs` files.
- **Unified error type**: Add `ProviderError` in `error.rs`, make `sign()` fallible across API handlers.
- **Testability**: Add helpers to `fs_index.rs` and `s3_index.rs` for test construction.
- **CI hardening**: Combine 3 pallet coverage jobs into one with glob discovery, add validation guards, fix integration test flakiness, fix `//Bob` nonce issue in S3 CI test.

## PR stack

|---|---|---|---|
| **1** | **This PR** | `dev` | Refactor + CI |
| 2 | #146  | `dev` | Pallet tests (independent) |
| 3 | #147  | This PR | HTTP integration tests |
| 4 | #148  | #147  | Coordinator + FS tests |

## Test plan

- [ ] `SKIP_WASM_BUILD=1 cargo clippy -p storage-provider-node --all-targets` passes
- [ ] CI check workflow passes
- [ ] No test files included — tests come in PRs 3 and 4

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -170,19 +170,39 @@ jobs:
 
       - name: Measure PR coverage
         run: |
-          PALLET_COV=$(cargo llvm-cov --lib -p pallet-storage-provider \
-            --ignore-filename-regex='(weights\.rs|runtime_api\.rs|mock\.rs|primitives/)' \
+          PALLET_COV=$(cargo llvm-cov --lib \
+            -p pallet-storage-provider \
+            -p pallet-drive-registry \
+            -p pallet-s3-registry \
+            --ignore-filename-regex='(weights\.rs|runtime_api\.rs|mock\.rs|primitives/|benchmarking\.rs|bechmarking\.rs)' \
             --summary-only 2>&1 \
             | grep '^TOTAL' | awk '{print $10}' | tr -d '%')
+          if [ -z "$PALLET_COV" ]; then
+            echo "ERROR: Failed to measure pallet coverage (build or test failure?)"
+            exit 1
+          fi
           echo "PR_PALLET_COV=$PALLET_COV" >> $GITHUB_ENV
           echo "PR pallet coverage: ${PALLET_COV}%"
 
           cargo llvm-cov clean --workspace 2>/dev/null || true
 
-          PROVIDER_COV=$(cargo llvm-cov --lib -p storage-provider-node \
-            --ignore-filename-regex='(\.cargo|rustc)' \
-            --summary-only 2>&1 \
-            | grep '^TOTAL' | awk '{print $10}' | tr -d '%')
+          # Discover available integration test targets from filesystem
+          TEST_FLAGS=""
+          for f in provider-node/tests/*_integration.rs; do
+            [ -f "$f" ] || continue
+            t=$(basename "$f" .rs)
+            TEST_FLAGS="$TEST_FLAGS --test $t"
+          done
+
+          if [ -n "$TEST_FLAGS" ]; then
+            PROVIDER_COV=$(cargo llvm-cov $TEST_FLAGS \
+              -p storage-provider-node \
+              --ignore-filename-regex='(\.cargo|rustc|_subxt\.rs|command\.rs|cli\.rs|main\.rs)' \
+              --summary-only 2>&1 \
+              | grep '^TOTAL' | awk '{print $10}' | tr -d '%')
+          else
+            PROVIDER_COV=0
+          fi
           echo "PR_PROVIDER_COV=$PROVIDER_COV" >> $GITHUB_ENV
           echo "PR provider coverage: ${PROVIDER_COV}%"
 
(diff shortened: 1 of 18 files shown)


.github/workflows/check.yml
```diff
@@ -170,19 +170,39 @@ jobs:
 
       - name: Measure PR coverage
         run: |
-          PALLET_COV=$(cargo llvm-cov --lib -p pallet-storage-provider \
-            --ignore-filename-regex='(weights\.rs|runtime_api\.rs|mock\.rs|primitives/)' \
+          PALLET_COV=$(cargo llvm-cov --lib \
+            -p pallet-storage-provider \
+            -p pallet-drive-registry \
+            -p pallet-s3-registry \
+            --ignore-filename-regex='(weights\.rs|runtime_api\.rs|mock\.rs|primitives/|benchmarking\.rs|bechmarking\.rs)' \
```
franciscoaguirre, line 177: bechmarking.rs?

storage-interfaces/s3/client/examples/ci_integration_test.rs
```diff
@@ -39,8 +39,11 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     println!();
 
     // Step 1: Create the S3 client
+    // Use //Bob (not //Alice) to avoid nonce contention with the provider node,
+    // which also runs as //Alice and submits background transactions
+    // (accept_agreement, checkpoints) via its coordinators.
```
ilchu, line 44: This nonce contention has been biting me forever in the UI tests btw. And this is what I did as well, offloading to Bob.

provider-node/Cargo.toml
```diff
@@ -33,3 +33,4 @@ codec = { workspace = true, features = ["std"] }
 [dev-dependencies]
 tokio = { workspace = true, features = ["macros", "rt-multi-thread", "time"] }
 serde_json = { workspace = true }
+tempfile = "3"
```
ilchu: I don't see where it's being used, am I missing something?

provider-node/src/challenge_responder_subxt.rs
```diff
@@ -0,0 +1,190 @@
...
+    ) -> Pin<Box<dyn Future<Output = Result<H256, Error>> + Send + '_>> {
+        Box::pin(async move {
+            // Build ChallengeId
+            let challenge_id_val = subxt::dynamic::Value::named_composite(vec![
+                (
+                    "deadline",
+                    subxt::dynamic::Value::u128(challenge_id.0 as u128),
+                ),
+                ("index", subxt::dynamic::Value::u128(challenge_id.1 as u128)),
+            ]);
```
ilchu: I believe we can rewrite all the dynamic value constructions (here and elsewhere) with subxt's `value!` [macro](https://docs.rs/subxt/latest/subxt/dynamic/macro.value.html). It would make it a bit more compact as maybe some type conversions could be automatic.

provider-node/src/agreement_coordinator.rs
```diff
@@ -5,35 +5,47 @@
...
-use subxt::{dynamic::Value, OnlineClient, PolkadotConfig};
-use subxt_signer::sr25519::Keypair;
+use storage_primitives::BucketId;
 use tokio::sync::mpsc;
 
+/// Trait abstracting chain interactions for the agreement coordinator.
+///
+/// Follows the same pattern as [`crate::auth::MembershipResolver`]: a trait
+/// with `Pin<Box<dyn Future>>` return types, stored as `Box<dyn AgreementChainClient>`,
+/// enabling mock-based testing without a live chain.
```
ilchu: Speaking about this: the membership resolver was built for mocking, but never got mocked / tested itself, now that I look at it. Or are you following it up in next PRs?

provider-node/src/agreement_coordinator.rs
```diff
@@ -5,35 +5,47 @@
...
+pub trait AgreementChainClient: Send + Sync {
+    /// Fetch bucket IDs with pending agreement requests for this provider.
+    fn fetch_pending_requests(
+        &self,
+        provider_account: &[u8; 32],
+    ) -> Pin<Box<dyn Future<Output = Result<Vec<BucketId>, Error>> + Send + '_>>;
```
ilchu: I'd prefer having more concise function signatures. This would avoid the type complexity lint disabling too. It can either be done with `#[async_trait::async_trait]` macro or rust-native return position impl Trait in trait. The latter is probably the best solution in terms of cleanness, and it would amount to something like this:
```rust
trait AgreementChainClient: Send + Sync {
    fn accept_agreement(&self, bucket_id: BucketId)
        -> impl Future<Output = Result<(), Error>> + Send;
}
```

and then in the actual impl:

```rust
impl AgreementChainClient for SubxtAgreementChainClient {
    async fn accept_agreement(&self, bucket_id: BucketId) -> Result<(), Error> {
        // ... async body, borrows &self fine ...
    }
}
```
Also I would lean to using impl trait wherever instead of boxed dyn trait objects, if possible.

provider-node/src/agreement_coordinator.rs
mudigal: Its surely getting written in next PRs.

storage-interfaces/s3/client/examples/ci_integration_test.rs
mudigal: PR #150 adds retry logic for Stale nonce errors globally in submitTx. This should somewhat fix this stale Nonce errors, hopefully all!

provider-node/Cargo.toml
mudigal: Good catch!

tempfile is used in disk_integration.rs which is in #147, and not in this PR. Missed it while splitting the PRs. Will move to #147.

ilchu (approved):

.github/workflows/check.yml
```diff
@@ -170,19 +170,39 @@ jobs:
...
+            t=$(basename "$f" .rs)
+            TEST_FLAGS="$TEST_FLAGS --test $t"
+          done
+
+          if [ -n "$TEST_FLAGS" ]; then
+            PROVIDER_COV=$(cargo llvm-cov $TEST_FLAGS \
+              -p storage-provider-node \
+              --ignore-filename-regex='(\.cargo|rustc|_subxt\.rs|command\.rs|cli\.rs|main\.rs)' \
+              --summary-only 2>&1 \
+              | grep '^TOTAL' | awk '{print $10}' | tr -d '%')
```
franciscoaguirre, line 202: I know it's unlikely to cause any issues, but can we add `--json` and use `jq` to extract the percentage. That way we need to fiddle less with text and it's easier to read

provider-node/src/checkpoint_coordinator_subxt.rs
```diff
@@ -0,0 +1,149 @@
...
+            sp_core::crypto::AccountId32::from(signer.public_key().0).to_ss58check()
+        );
+        tracing::info!("Checkpoint coordinator connected to {}", chain_ws_url);
+
+        Ok(Self { api, signer })
+    }
+}
+
+#[async_trait::async_trait]
+impl CheckpointChainClient for SubxtCheckpointChainClient {
```
franciscoaguirre, line 38: Feels like all these different traits could be implemented by the same struct to reuse the subxt api and signer. We can still keep the impls in different files

franciscoaguirre (commented):
What do you think about having the same struct implement all the different chain client traits? Would it reuse connections?

PR #146, opened 2026-06-08
test(pallet): add tests for challenges, checkpoints, replicas, events, and runtime API

mudigal:
## Summary

#144 

Split from #139 (PR 2 of 4). Independent of the other PRs — pallet-only changes.

Add comprehensive test coverage for the storage-provider pallet:

- **Challenge tests**: cost-split boundary conditions, offchain challenge flows
- **Checkpoint tests**: scheduling, fallback, leader election, missed checkpoint handling
- **Misc tests**: event emission, `remove_slashed`, multiaddr validation
- **Replica tests**: full replica sync lifecycle
- **Runtime API tests**: provider/bucket/agreement query coverage

## PR stack

|---|---|---|---|
| 1 | #145 | `dev` | Refactor + CI |
| **2** | **This PR** | `dev` | **Pallet tests (independent)** |
| 3 | test/provider-node-http-tests | #145 | HTTP integration tests |
| 4 | test/coordinator-and-fs-pallet-tests | PR 3 | Coordinator + FS tests |

## Test plan

- [ ] `cargo test -p pallet-storage-provider` passes
- [ ] No runtime or production code changes — pure test additions

### DIFF
--- a/pallet/src/mock.rs
+++ b/pallet/src/mock.rs
@@ -127,7 +127,6 @@ pub fn new_test_ext() -> sp_io::TestExternalities {
 }
 
 /// Build test externalities with custom balances.
-#[allow(dead_code)]
 pub fn new_test_ext_with_balances(balances: Vec<(u64, u64)>) -> sp_io::TestExternalities {
     let mut t = frame_system::GenesisConfig::<Test>::default()
         .build_storage()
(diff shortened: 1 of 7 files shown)


pallet/src/mock.rs
```diff
@@ -144,6 +144,11 @@ pub fn new_test_ext_with_balances(balances: Vec<(u64, u64)>) -> sp_io::TestExter
 }
 
 /// Run to a specific block number.
+///
+/// **WARNING**: This only calls `System` hooks, NOT `StorageProvider::on_finalize`.
+/// This means challenge expirations and slashing will NOT be processed automatically.
+/// If your test needs to trigger challenge timeout slashing, call the pallet hook
+/// manually: `<StorageProvider as Hooks<u64>>::on_finalize(block_number);`
 #[allow(dead_code)]
```
ilchu: It's pretty much a nit because not PR related, but I believe all the allows are redundant in this file. And besides the `test_public_key` is repeated in `pallet-s3-registry/src/tests.rs` **and** `pallet/src/tests/mod.rs`.. 😅

pallet/src/mock.rs
```diff
@@ -144,6 +144,11 @@ pub fn new_test_ext_with_balances(balances: Vec<(u64, u64)>) -> sp_io::TestExter
 }
 
 /// Run to a specific block number.
+///
+/// **WARNING**: This only calls `System` hooks, NOT `StorageProvider::on_finalize`.
+/// This means challenge expirations and slashing will NOT be processed automatically.
+/// If your test needs to trigger challenge timeout slashing, call the pallet hook
+/// manually: `<StorageProvider as Hooks<u64>>::on_finalize(block_number);`
```
ilchu: IMO our pallet's custom slashing hook should be included into the mock `run_to_block` call as well. That's what happens in runtime too, as all participating pallets' hooks are called. And it would only break `respond_to_challenge_fails_expired` AFAICT, but that test is already testing a scenario that should be unreachable anyways.

ilchu: Also as another side nit: just to be safe we should be checking `if block_number() > 1` in this fn as well -- because the custom hook should never run on genesis block.

ilchu (commented):
Mostly nits and ergonomics here as well, but pretty much LGTM.

pallet/src/mock.rs
mudigal: Good catch on the redundant allows - removed them. The test_public_key duplication is a pre-existing issue across crates; I'll leave that for a separate cleanup to keep this PR scoped to pallet tests.

pallet/src/mock.rs
mudigal: > IMO our pallet's custom slashing hook should be included into the mock `run_to_block` call as well. That's what happens in runtime too, as all participating pallets' hooks are called. And it would only break `respond_to_challenge_fails_expired` AFAICT, but that test is already testing a scenario that should be unreachable anyways.

Agreed — the mock should match runtime behavior. Updated `run_to_block` to call `StorageProvider::on_finalize` on each block. This broke `respond_to_challenge_fails_expired` as you predicted - removed it since it was testing an unreachable scenario (responding after the deadline, which `on_finalize` would have already slashed).


> Also as another side nit: just to be safe we should be checking if block_number() > 1 in this fn as well -- because the custom hook should never run on genesis block.

Done — the hook is guarded with `if i > 1` to skip genesis.

ilchu (approved):

PR #147, opened 2026-06-08
test(provider-node): add HTTP and storage integration tests

mudigal:
## Summary

#144 

Split from #139 (PR 3 of 4). Stacked on #145.

Add integration tests for the provider node's HTTP API layer:

- **API integration**: All HTTP endpoints (upload, download, commit, checkpoint, proof, delete, etc.)
- **Auth integration**: Authentication middleware and signature verification
- **Disk integration**: On-disk storage backend operations
- **FS integration**: File-system API endpoint tests
- **S3 integration**: S3-compatible API endpoint tests

## PR stack

|---|---|---|---|
| 1 | #145 | `dev` | Refactor + CI |
| 2 | #146 | `dev` | Pallet tests (independent) |
| **3** | **This PR** | **#145** | **HTTP integration tests** |
| 4 | test/coordinator-and-fs-pallet-tests | This PR | Coordinator + FS tests |

## Test plan

- [ ] `SKIP_WASM_BUILD=1 cargo clippy -p storage-provider-node --all-targets` passes
- [ ] `cargo test -p storage-provider-node` — HTTP/storage tests pass
- [ ] Pure test additions — no production code changes


provider-node/tests/disk_integration.rs
```diff
@@ -0,0 +1,422 @@
...
+        let disk = DiskStorage::new(dir.path()).expect("RocksDB should open");
+        let state = Arc::new(
+            ProviderState::with_seed(Arc::new(disk), "//Alice").expect("//Alice is valid"),
+        );
+
+        let app = create_router(state);
+        let listener = TcpListener::bind("127.0.0.1:0").await.unwrap();
+        let addr = listener.local_addr().unwrap();
+        tokio::spawn(async move { axum::serve(listener, app).await.unwrap() });
+        tokio::time::sleep(std::time::Duration::from_millis(10)).await;
```
ilchu: Minor, but I believe the sleep is not needed as any incoming requests would still queue healthily upon spawn.

provider-node/tests/auth_integration.rs
```diff
@@ -0,0 +1,462 @@
...
+        }
+    }
+}
+
+impl MembershipResolver for MockResolver {
+    fn fetch_members(
+        &self,
+        _bucket_id: u64,
+    ) -> std::pin::Pin<
+        Box<dyn std::future::Future<Output = Result<Vec<(AccountId32, Role)>, String>> + Send + '_>,
```
ilchu: Can we apply the same suggestion here as before with the `[async_trait::async_trait]` for the compact return sig?

provider-node/Cargo.toml
```diff
@@ -34,3 +34,4 @@ async-trait = "0.1"
 [dev-dependencies]
 tokio = { workspace = true, features = ["macros", "rt-multi-thread", "time"] }
 serde_json = { workspace = true }
+tempfile = "3"
```
ilchu, line 37: Sorry for coming back to it again, but I've just noticed this and some other dependencies are not workspaced properly. It would be a bit more systematic to have them declared in the root `Cargo.toml` and inherit the workspace version like other deps do. But this also applies to some untouched code, so I'm just flagging.

ilchu (approved):
Once again, just ergonomics here.

provider-node/tests/auth_integration.rs
mudigal: Done, converted `MembershipResolver` trait and both impls (`ChainMembershipResolver`, `MockResolver`) to `#[async_trait]`.

provider-node/Cargo.toml
mudigal: Agreed, this applies to some pre-existing deps too. Tracked in #154 to address in a dedicated cleanup pass.

ilchu (approved):

franciscoaguirre (approved):

provider-node/src/auth.rs
```diff
@@ -119,75 +114,66 @@ impl ChainMembershipResolver {
...
-                        if let Some(bytes) = account_bytes {
-                            let account = AccountId32::from(bytes);
-                            let role = extract_role(role_v);
-                            members.push((account, role));
-                        }
+    async fn fetch_members(&self, bucket_id: u64) -> Result<Vec<(AccountId32, Role)>, String> {
+        use subxt::dynamic::At;
+        use subxt::{dynamic::Value, OnlineClient, PolkadotConfig};
+
+        let api = OnlineClient::<PolkadotConfig>::from_url(&self.chain_rpc)
```
bkontur, line 123: @franciscoaguirre is this the reason why you started https://github.com/paritytech/web3-storage/pull/159? 

From what I reading about subxt:
```
  - new() / from_url() are the expensive parts (network handshake + metadata fetch). Call them once at startup, not per request.
  - Don't reconnect on every operation. A common anti-pattern is calling from_url inside a loop or per request — that opens a fresh WS connection and re-downloads metadata each time.

  A quick rule of thumb: treat OnlineClient<PolkadotConfig> like an Arc-backed handle (similar to a reqwest::Client or a DB pool) — construct once, clone everywhere.
```

@franciscoaguirre will you please check and refactor all this `OnlineClient::<PolkadotConfig>` also as a part of your https://github.com/paritytech/web3-storage/pull/159 (or separate PR)?

bkontur:
Looks like some flaky test was introduced here, all the PRs on the top of this fails Basic tests. Potential fix introduced here: https://github.com/paritytech/web3-storage/pull/161

PR #151, opened 2026-06-09
fix(ci): pass explicit best-block nonce for SC contract txs

x3c41a:
Builds on #150. The stale-nonce race surfaces on more than the contract path: CI hit it on the provider account (`//Alice`) during setup, which the provider node's coordinators sign with too. Two refinements to the retry:

- Apply it to `submitTxFinalized` as well as `submitTx` (the challenge paths use finalized submits and can stale the same way).
- On retry, drop any explicit nonce so PAPI re-reads a fresh one, and match `Stale` directly rather than via a message-format-specific regex.

I first tried passing an explicit best-block nonce on the contract calls, but that could read a nonce already ahead of the validating context, leaving the tx in Future and timing out after 180s (a `Revive.call` timeout on web3-storage-paseo) — and the Stale retry can't recover from that. Reverted; the retry alone covers the race.

Verified locally (omni-node dev chain + inmemory provider, web3-storage-paseo runtime): the full CI sequence `sc-flow` → `sc-coverage` → `sc-team-drive` → `sc-token-gated` with drain between, 3 iterations, 12/12 pass. Also forced a real "nonce too low" stale and confirmed `submitTx` retried with a fresh nonce and the tx landed.

### DIFF
--- a/examples/papi/common.js
+++ b/examples/papi/common.js
@@ -346,6 +346,41 @@ export async function waitForTransaction(
   });
 }
 
+// Retry budget for `Invalid::Stale` (nonce too low). Stale happens two ways on
+// these dev chains: a fresh PAPI client whose nonce view trails the chain, and
+// the provider account racing the provider node's coordinators, which sign from
+// the same key. Contract calls pass an explicit best-block nonce to avoid the
+// first case deterministically; this retry is the backstop for the race, which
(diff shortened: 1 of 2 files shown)


x3c41a:
I thought I'd find an approach better than retrying but I did not manage to. Closing as duplicate

PR #155, opened 2026-06-10
fix: update source of cmd bot

danielbui12:
Using protected `cmd-bot` branch for `cmd.py`.

> "The protected cmd-bot branch allows to check out `cmd-run.yml` from a trusted source so users can't modify it (as I described in this comment) and at the same time the scope of the branch is narrower than main branch."

Ref: https://github.com/paritytech/security/issues/103


bkontur (approved):

ilchu (approved):
I tried asking the SRE if we need to make the branch, and they said yes, so it was there. But funnily enough it was still running off of dev.

PR #159, opened 2026-06-11
Reuse one SubxtChainClient struct that implements all chain client traits in provider-node

franciscoaguirre:
Keeps the separation of the interfaces but deduplicates code


provider-node/src/subxt_client.rs
```diff
@@ -206,15 +238,227 @@ impl SubxtReplicaSyncChainClient {
...
+                };
+
+                tracing::info!(
+                    "Found pending agreement request for us: bucket {}",
+                    bucket_id
+                );
+                bucket_ids.push(bucket_id);
+            }
+
+            if entry_count > 0 {
```
bkontur: hmm, why not use just `bucket_ids.len()` here?

`entry_count` is confusing, it is incremented at the beginning between various `continue;`, so it is not obvious what `entry_count` means - success, correct, all, partial?

bkontur:
@franciscoaguirre please, check also this comment: https://github.com/paritytech/web3-storage/pull/147/changes#r3394340663 - we should check and refactor all the `OnlineConfig` usages

franciscoaguirre:
@bkontur I wanted to keep this PR small and just do this one refactor but I can create a PR on top that tries to reuse client instantiation

provider-node/src/subxt_client.rs
franciscoaguirre: I moved this exactly as it was before in order to not change the behavior in any way. I'll make follow-up PRs simplifying the functions themselves

provider-node/src/lib.rs
danielbui12: Question: Why do we use `sp_core::sr25519::Pair` here but in challenge_responder, agreemeent_coordinator, replica_sync_coordinator use `subxt_signer::sr25519::Keypair` ?

I believe we can refactor all to use `subxt_signer::sr25519::Keypair` so that there will not `pub async fn connect(chain_ws_url: &str, seed: &str)` or `pub async fn connect(chain_ws_url: &str, keypair: &sr25519::Pair)`

bkontur: > Question: Why do we use `sp_core::sr25519::Pair` here but in challenge_responder, agreemeent_coordinator, replica_sync_coordinator use `subxt_signer::sr25519::Keypair` ?
> 
> I believe we can refactor all to use `subxt_signer::sr25519::Keypair` so that there will not `pub async fn connect(chain_ws_url: &str, seed: &str)` or `pub async fn connect(chain_ws_url: &str, keypair: &sr25519::Pair)`

@danielbui12 please, follow up

provider-node/src/subxt_client.rs
bkontur: > I moved this exactly as it was before in order to not change the behavior in any way. I'll make follow-up PRs simplifying the functions themselves

nice, thank you, I want to merge this, so I just created follow-up here: https://github.com/paritytech/web3-storage/issues/174 for this

bkontur (approved):

provider-node/src/lib.rs
danielbui12: @bkontur I remembered that it is already fixed somewhere https://github.com/paritytech/web3-storage/blob/dev/provider-node/src/lib.rs#L71C1-L72C1

mudigal:
I would appreciate if code coverage was maintained, if not increased. That is the reason that check was introduced in Basic Check.

Issue #136, opened 2026-06-04
Extract shared chain-waiting utilities to reduce E2E and Playwright test duration/flakiness


mudigal:
## Context

Raised in PR #129 [review comment](https://github.com/paritytech/web3-storage/pull/129#discussion_r3356354779).

The polling-based wait helpers in `examples/papi/common.js` (`waitForBlock`, `waitForAgreementAcceptance`, etc.) are a source of slowness and potential flakiness in both the PAPI E2E test suite and the UI Playwright E2E tests.

## Problem

- **Polling is slow**: `waitForAgreementAcceptance` polls chain state every few seconds with a 60s timeout. With 2s dev-mode blocks, this adds unnecessary latency.
- **Shared pain**: The UI Playwright tests face the same issue — they also need to wait for on-chain state changes (agreement accepted, bucket created, etc.) and likely use similar polling patterns.
- **Flakiness risk**: Polling with fixed timeouts can flake under CI load. Event-based waiting would be more deterministic.

## Proposed improvements

1. **Event-based waiting** — subscribe to finalized block events and filter for the expected pallet event (e.g. `StorageProvider::AgreementAccepted`) instead of polling storage. Faster and more reliable.
2. **Shorter block times for Playwright** — if Playwright tests don't already use `--dev-block-time`, switching from Zombienet (6s blocks + finality lag) to dev mode (2s instant-finalized blocks) would cut wait times significantly.
3. **Shared test-helpers package** — extract `waitForBlock`, `waitForAgreementAcceptance`, and similar utilities into a shared package that both PAPI E2E and Playwright tests consume (related to #135).

## Acceptance criteria

- [ ] Audit current wait patterns in E2E and Playwright tests
- [ ] Implement event-subscription-based waiting where possible
- [ ] Evaluate dev-mode chain for Playwright tests
- [ ] Measure before/after test suite duration

Issue #123, opened 2026-06-03
Create a dedicated TS SDK


danielbui12:
## Context

We have a three-layer storage architecture:

- **Layer 0** — the storage primitive.
- **Layer 1** — two interchangeable storage interfaces **driver storage** and **S3 storage**.
- **Layer 2** — user interfaces.

The main goal is to build a TypeScript SDK that can be reusable across all package (examples, UI, client):, a lot of logics (content addressing, chunking, crypto, retries, error handling) are identical.

This issue proposes a structure to get there and lists the decisions we need to make before writing code.

## Proposed package layout

A monorepo (pnpm) with a backend-free `core` package. The physical separation is the point: it forces shared code to actually be shared rather than quietly coupling to S3 or the chain.

```
web3-storage/
  packages/
    core/
      src/
        interfaces.ts
        errors.ts
        codec/
        crypto/
        stream/
        utils/

    layer0/             # the storage primitive binding

    layer1/              # Layer 1: the two backends
      src/
        driver-registry/
        s3-registry/
        shared/           # logic common to both but not core-worthy
```

Dependency direction is strictly downward: `layer1` -> `layer0` -> `core`. Nothing points back up.

## Tasks

- [ ] ...

bkontur:
one issue for SDK is enough, closing and continue in https://github.com/paritytech/web3-storage/issues/64

PR #160, opened 2026-06-11
Transition to TS SDK

### LINKED ISSUE #135: Unify PAPI helper wrappers across E2E tests, SC examples, and UIs
## Context

Raised in PR #129 [review comment](https://github.com/paritytech/web3-storage/pull/129#discussion_r3356329429).

`examples/papi/api.js` contains thin wrappers around pallet extrinsics (`createBucket`, `requestPrimaryAgreement`, `endAgreement`, etc.) that are very similar to wrappers in the UI layer (`provider.state.ts`) and SC example scripts.

## Problem

The same pallet interaction patterns are duplicated in multiple places:
- `examples/papi/api.js` — E2E test helpers
- `examples/papi/common.js` — shared utilities (signer, address comparison, polling)
- `user-interfaces/shared/test-helpers/` — Playwright test helpers
- `user-interfaces/provider/src/state/` — UI state layer
- SC example scripts

Bug fixes and API changes (e.g. new extrinsic fields, enum encoding) must be applied in every copy.

## Options to explore

1. **Shared npm workspace package** (e.g. `@web3-storage/papi-helpers`) — extract common wrappers into a single package. Both E2E tests and UIs import from it.
2. **Shared types/codegen only** — share the PAPI-generated descriptors (`.papi/`) but keep wrapper logic separate, since E2E (throw-on-failure, log finalization) and UI (error handling, state updates) have different concerns.
3. **Keep separate, add lint** — accept the duplication but add a CI check that flags drift between the copies.

## Acceptance criteria

- [ ] Decide on approach
- [ ] Implement shared package or document why duplication is intentional
- [ ] Ensure E2E tests, SC examples, and UIs all consume the shared code (or document exceptions)

ilchu:
Closes #135 and #136; realises the SDK direction from #123.

## What

Extracts a single `@web3-storage/sdk` that every chain consumer imports — the PAPI e2e suite, the smart-contract demos, all three UIs (drive-ui + provider + s3-ui), and the Playwright test-helpers — replacing the hand-maintained copies of the pallet wrappers, dev-signer derivation, and chain-waiting helpers that #135 flagged. With s3-ui migrated too (see "s3-ui on the SDK" below), **no hand-rolled chain/PAPI helper copies remain anywhere — #135 is fully closed.** The wait helpers are `watchValue`-based from the start rather than polling, which is the #136 half (no more 1–1.5s poll loops dominating test wall-clock). The pnpm workspace is hoisted to the repo root, so the whole repo resolves one lockfile, one generated descriptor set, and one `polkadot-api` instance.

Package layout follows #123's monorepo shape: `core` (backend-free: http/fetch+retry, base64, CID verify, byte/address utils) → `layer0` (client/signers/tx/waits, pallet wrappers, provider-http, revive) → `layer1` (FileSystemClient + S3Client) → `sdk` (umbrella re-export + the `Web3Storage` facade). Dependency direction is strictly `layer1 → layer0 → core`.

## Rebased onto dev's off-chain agreements (#105) + deps bump (#173)

dev moved agreement negotiation off-chain in #105: a provider signs `AgreementTerms` out-of-band (`POST /negotiate`) and the owner redeems them on-chain in one atomic extrinsic (`establish_storage_agreement`, or `create_drive` / `create_s3_bucket` for Layer 1) — the old `create_bucket` + `request_agreement` + `accept_agreement` dance is gone. This branch adopts that flow entirely through the SDK: `negotiateTerms` + `buildSignedTermsArgs` + `establishStorageAgreement`/`establishReplicaAgreement` live in layer0; `FileSystemClient.createDrive` and `S3Client.createBucket` resolve a provider (auto-discovered, or explicit) → negotiate → redeem, with a pre-negotiated `SignedTerms` override; the examples, the UIs (via the provider-picker dialog), and the test-helpers all route through it. #173's bump (`polkadot-api` 2.1.6, `@polkadot-api/substrate-bindings` 0.20.3, `typescript` 6.0.3, plus react/vite/vitest/playwright) is reconciled into the single root manifest set.

## Caught up to dev: S3-UI series, console-ui removal, provider challenge, v0.3.0 + repo hygiene

dev replaced the old console-ui with a dedicated `s3-ui` app (#182–#186), removed console-ui outright (#188), added a manual challenge-response page to the provider UI (#189), then shipped v0.3.0-paseo alongside a repo-wide SPDX-header sweep (#201), LICENSE files + a GPL-3.0-only workspace license (#199), a local-testnet runtime move to `runtimes/web3-storage-local` (#204), and CONTRIBUTING / sub-crate READMEs / security-hygiene (#200/#202/#203). This branch absorbs all of it via merge: console-ui is dropped (its remaining value was the SDK packages plus drive-ui/provider/examples/test-helpers, which stay); `s3-ui` is brought in as a root-workspace member wired into the `ui-checks` matrix + `deploy-ui`; the provider's `chain-client` keeps its SDK-backed `submit`/`submitTx` path while gaining #189's `fetchChallengeProof` + `submitRespondToChallenge` on top of it. The paseo bumps (v0.2.0 → v0.3.0) only move `spec_version`/`transaction_version`, so the PAPI descriptors are unaffected (regenerate to a no-op). New-on-branch TS sources (the SDK packages, `examples/papi`, the new test-helpers) carry SPDX headers matching dev's #201 convention — `GPL-3.0-only`, except `examples/papi` which stays `Apache-2.0` like its manifest.

## s3-ui on the SDK (closes the last #135 gap)

s3-ui's 884-line hand-rolled `S3Client` — a re-implementation of object HTTP, CID verification, `/negotiate`, MultiSignature encoding, pallet-tx building, provider-URL resolution, and request signing — is replaced by a thin adapter over the SDK, mirroring drive-ui's `drive-client.ts`: bucket/object ops delegate to `@web3-storage/sdk/s3`'s `S3Client`, member writes to layer0 `setMember`/`removeMember`, the challenge write to the SDK's `submitTx` + `requireOneEvent`; `negotiateTerms` / `buildSignedTermsArgs` / `parseMultiaddrToUrl` / `httpFetch` are re-exported from the SDK rather than duplicated; `crypto.ts` is deleted (`seedToKeypair` / `toSs58` / `Keypair` come from the SDK). Only app-shaped concerns stay local — provider discovery via the `find_matching_providers` runtime API, the UI view types, and the member/checkpoint/challenge reads + subscriptions — exactly as in drive-ui. The file shrinks 884 → 650 lines; s3-ui now imports `@web3-storage/sdk` and **zero** `@web3-storage/papi`.

## CI

The metadata drift check regenerates descriptors against the live chain and fails on any drift; on failure it now uploads the freshly-generated `parachain.scale` as a `fresh-papi-metadata` artifact and prints the exact one-command fix, because SCALE metadata is not byte-reproducible across build environments (so a runtime author adopts CI's bytes rather than regenerating locally). The existing `integration-tests` workflow continues to gate the PAPI e2e suite (with c8 coverage over the sdk + examples), the smart-contract demos, the zombienet smoke test, and the UI Playwright suites (drive-ui + provider) — all of which exercise the negotiate→establish flow end-to-end.

## Also

Removes the orphaned `user-interfaces/sdk/typescript` stub (zero consumers, never built/tested, violated the `@polkadot/*` ban) and repoints the docs at `@web3-storage/sdk`. The `Web3Storage.connect()` facade wraps the flat functions additively, so #123's class API can grow on top without reworking callers.


user-interfaces/shared/papi/src/index.ts
```diff
@@ -102,7 +104,7 @@ export async function resolveProviderEndpoint(
   }
 
   for (const providerAccount of providers) {
-    const provider = await api.query.StorageProvider.Providers.getValue(providerAccount);
+    const provider = await api.query.StorageProvider.Providers.getValue(providerAccount, { at: "best" });
```
bkontur: @ilchu I introduced for `examples` `constant export const READ_OPTS = { at: "best" };`, but for real usage, not sure if BestBlock is ok, maybe we should stick to the finalized (explicitly -> refactor to some constant, so we can change easily)

user-interfaces/console-ui/src/lib/storage.ts
```diff
@@ -459,14 +378,15 @@ export class StorageClient {
...
+      this.signerAddress!,
+      READ_OPTS,
     );
 
     if (!bucketIds) return [];
 
     const buckets: BucketInfo[] = [];
     for (const bucketId of bucketIds) {
-      const bucket = await this.api!.query.S3Registry.S3Buckets.getValue(bucketId);
+      const bucket = await this.api!.query.S3Registry.S3Buckets.getValue(bucketId, READ_OPTS);
```
bkontur: @ilchu for real UI, probably we should go with finalize, as I said before better refactor this and set the separate const as finalized for now

bkontur (approved):
lgtm modulo I would keep the real UIs using "finalized":
- either remove READ_OPTS from UIs
- or introduce separate const for UIs READ_OPTS = "finalized-or-what-is-the-format"

we can keep READ_OPTS = "at: best" for examples and tests, that is ok

https://github.com/paritytech/web3-storage/pull/160#discussion_r3395443374

bkontur (changes requested):
https://github.com/paritytech/web3-storage/pull/160#pullrequestreview-4476262539

bkontur:
@ilchu please, try to rebase :)

bkontur (approved):

packages/papi/.papi/polkadot-api.json
```diff
@@ -5,8 +5,8 @@
     "parachain": {
       "wsUrl": "ws://127.0.0.1:2222",
       "metadata": ".papi/metadata/parachain.scale",
-      "genesis": "0x8ae30e626ea89c4a00492b8f8c36dfdace1e987adef42759fd1fdb9aec1f2e3e",
-      "codeHash": "0xc4028d537da65cea73c8a24e676bfa3fd3617b8563a848c191c044294ebbdff5"
+      "genesis": "0x4545454545454545454545454545454545454545454545454545454545454545",
```
bkontur, line 8: @ilchu do we need to push this anytime the runtime changes?

ilchu: Yeah it's probably given more grief than use by now.

PR #163, opened 2026-06-11
fix: revert provider deregistration

### LINKED ISSUE #107: Investigate optimal provider deregistration

danielbui12:
## Changes

- committed_bytes == 0 is the authoritative gate, proven faithful by auditing every StorageAgreements insert/remove site; each is balanced against a counter increment/decrement. No provider→bucket reverse index needed.
- remove cancel_deregistration, complete_deregistration

## Issue
- Close: https://github.com/paritytech/web3-storage/issues/107


danielbui12:
/cmd bench --runtime web3-storage-paseo --runtime storage-parachain-runtime

danielbui12:
/cmd bench --runtime web3-storage-paseo --runtime storage-parachain-runtime

PR #158, opened 2026-06-11
Harden GitHub Actions with zizmor

franciscoaguirre:
Audited all github workflows with zizmor and fixed a bunch of issues. 179 findings -> 0 (17 documented exceptions).

- /cmd bot: org-members-only (dropped PR-author path), scoped the GitHub App tokens (members: read / contents: write, were full-installation), least-privilege per job.
- Minimal permissions: on all workflows; deploy-ui pages/id-token moved to its deploy job.
- Template-injection closed via env-indirection (workflows + 4 composite actions).
- persist-credentials: false on non-pushing checkouts; dependabot cooldown.
- New zizmor.yml CI gate + .github/zizmor.yml (keeps this from regressing).

PR #167, opened 2026-06-11
Make is_pr_author optional in /cmd

franciscoaguirre:
The /cmd flow dispatches cmd-run.yml with `gh workflow run --ref cmd-bot`, so workflow_dispatch inputs are validated against this branch. After the PR-author authorization path was dropped from cmd.yml (in #158), it stopped sending is_pr_author, which 422'd the dispatch ("Required input 'is_pr_author' not provided") because the input was required here.

Make it optional with a default and drop the now-dead IS_PR_AUTHOR env var (the Run cmd step never read it). Keeping the input declared rather than removing it avoids an "Unexpected inputs provided" 422 from cmd.yml copies that still pass the flag, so the two branches can be updated in any order.

### DIFF
--- a/.github/workflows/cmd-run.yml
+++ b/.github/workflows/cmd-run.yml
@@ -21,10 +21,6 @@ on:
       is_org_member:
         description: "Is the user an org member"
         required: true
-      is_pr_author:
-        description: "Is the user the PR author"
-        required: false
-        default: "false"
       repo:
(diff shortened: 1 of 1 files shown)


franciscoaguirre:
@danielbui12 take a look

danielbui12 (approved):

PR #170, opened 2026-06-12
Move design from polkadot-sdk

eskimor:
Plus some clarification on challenger rewards + PDP and Filecoin corrections.

This is now the canonical doc.


docs/design/scalable-web3-storage.md
```diff
@@ -37,55 +64,139 @@
...
+But PDP still requires scheduled proofs—one per data set per day. At network
+scale with millions of data sets, the chain must still process millions of proof
+submissions daily. **Chain throughput still bounds storage capacity**, just at a
+higher ceiling. Write latency is still bound by chain latency.
 
-Both approaches struggle with the same question: **How do we know data is available?**
+### The Deeper Problem: What Do Proofs Actually Guarantee?
 
-Filecoin answers with cryptography: continuous proofs. This is expensive and doesn't scale.
+Filecoins proving ensures at a high cost that data is stored, but there is no
```
franciscoaguirre:
```suggestion
Filecoin's proving ensures at a high cost that data is stored, but there is no
```

docs/design/scalable-web3-storage.md
```diff
@@ -356,79 +531,156 @@ At agreement end, the owner decides: pay the provider, or burn the locked paymen
...
+**Service tiers emerge naturally:**
+1. **Anonymous** (no Proof-of-DOT): Served only when spare capacity exists, no reputation tracking
+2. **Registered** (Proof-of-DOT): Always get basic service, reputation tracked, can build payment history
+3. **Paying** (Proof-of-DOT + payments): Premium service based on payment history
 
-Competition drives quality. The feedback loop is natural:
+**Distinction from Proof of Personhood:** Proof-of-DOT allows multiple identities per person (if they pay for each).
+It's designed for abundant resources (bandwidth, connections) where we want sustainable economics and fast verification
+(network level). Proof of Personhood is for truly scarce resources (votes, airdrops) where one-per-human matters
+(blockchain level). They complement each other: proven persons could get DOT for Proof-of-DOT registration for free.
```
franciscoaguirre:
```suggestion
(blockchain level). They complement each other: proven people could get DOT for Proof-of-DOT registration for free.
```

docs/design/scalable-web3-storage.md
```diff
@@ -356,79 +531,156 @@ At agreement end, the owner decides: pay the provider, or burn the locked paymen
...
+### How Competition Drives Quality
 
-Clients connect to multiple providers, experience service quality directly, and vote with their feet. No complex monitoring required—just "did this work well for me?"
+The feedback loop is natural:
+1. Clients experience service quality directly
+2. Good service → continued payments
+3. Poor service → switch providers or stop paying
+4. Providers compete for paying clients
+
+Example: A viral video receives 10,000 requests. Paying users stream instantly, free users buffer. Client software
```
franciscoaguirre, line 670: The idea is that if the free user buffers and they don't like that then creating a challenge is more expensive than just paying to get better QoS?

franciscoaguirre (approved):

docs/design/scalable-web3-storage.md
eskimor: Challenging is definitely more expensive and not necessarily effective. A standard implementation will just serve the challenge, but don't improve service otherwise. The model (which should also be game theoretically incentivized) is, challenge if you can't get the data at all, burn if you just have been served really poorly. That being said, a challenge is also a very expensive and slow way of recovering data, which may be a useful escape hatch to recover the most crucial pieces, if things go south. (Malicious provider, no backup, ...), but that's it.

A challenge would be something you might consider if effective data rates are so low, that you can't effectively recover the data and what could happen is that providers which have something to lose, is that the ymonitor challenges and if you keep challenging (to really recover data), they might indeed change their setup to better serve you directly, instead of keep playing that costly challenge game. But that's not something that will happen because of a single challenge .... so yes definitely more costly.

bkontur (approved):

bkontur:
@eskimor Thank you, merging now.

I'm also considering adding guidance to CLAUDE.md, or creating a dedicated skill, that can verify on demand whether the implementation matches the canonical design and highlight any divergences.

E.g.: https://github.com/paritytech/web3-storage/issues/177

eskimor:
> I'm also considering adding guidance to CLAUDE.md, or creating a dedicated skill, that can verify on demand whether the implementation matches the canonical design and highlight any divergences.

Good idea!

mudigal:
Lets not force merge when Cargo Clippy is failing.

PR #172, opened 2026-06-13
fix: revert provider deregistration

### LINKED ISSUE #107: Investigate optimal provider deregistration

danielbui12:
## Changes

- committed_bytes == 0 is the authoritative gate, proven faithful by auditing every StorageAgreements insert/remove site; each is balanced against a counter increment/decrement. No provider→bucket reverse index needed.
- remove cancel_deregistration, complete_deregistration

## Issue
- Close: https://github.com/paritytech/web3-storage/issues/107


danielbui12:
/cmd bench --runtime web3-storage-paseo --runtime storage-parachain-runtime --pallet pallet-storage-provider --pallet pallet-s3-registry --pallet pallet-drive-registry

danielbui12:
/cmd bench --runtime web3-storage-paseo --runtime storage-parachain-runtime --pallet pallet_storage_provider pallet_s3_registry pallet_drive_registry

.github/workflows/integration-tests.yml
```diff
@@ -298,18 +295,39 @@ jobs:
...
+          url: http://127.0.0.1:3333
+          label: alice
+          log-file: /tmp/provider-alice.log
+
+      - name: Start provider (charlie - inmemory)
+        run: |
+          nohup ./target/release/storage-provider-node \
+            --keyfile /tmp/charlie-key --storage-mode inmemory \
+            --bind-addr 0.0.0.0:3334 --chain-rpc ws://127.0.0.1:2222 \
+            --enable-checkpoint-coordinator > /tmp/provider-charlie.log 2>&1 &
```
ilchu, line 323: Considering that we're repeating this one as well now, maybe we should make an action for this just like the `wait-for-provider-health`?

examples/papi/e2e/01-provider-registration.js
```diff
@@ -251,6 +252,7 @@ async function main() {
         });
         assert.ok(entry, "Charlie should still be listed");
         assert.strictEqual(entry.match_score, 0, "Non-accepting provider should score 0");
+        assert.strictEqual(entry.info.accepting_primary, false, "Non-accepting primary provider should be false");
```
ilchu:
```suggestion
        assert.strictEqual(entry.info.accepting_primary, false, "Non-accepting primary provider should have the corresponding field unset");
```

ilchu (approved):
LGTM besides two nits. Nice drive-by coverage bump too.

.github/workflows/integration-tests.yml
danielbui12: the CI need to be improved over times 😂; but I prefer fixing in another PR

mudigal:
## Concern: slashing race when removing the two-step deregistration

The old two-step flow (`deregister_provider` → wait `DeregisterAnnouncementPeriod` → `complete_deregister`) existed to close a slashing race. With the single-step approach, I believe this race is reintroduced:

1. Provider has an active agreement (`committed_bytes > 0`)
2. A challenger creates a challenge against the provider (stored in `Challenges` storage)
3. The agreement ends/expires → `committed_bytes` drops to 0
4. Provider immediately calls `deregister_provider` → stake is unreserved and returned
5. The challenge deadline hasn't passed yet — but the provider's stake is already gone, so slashing has no effect

The `committed_bytes == 0` check only guards against active **agreements**, but pending **challenges** are stored separately and can outlive the agreement they were created against.

### Possible fixes

- Add a check inside `deregister_provider` that ensures no pending challenges exist for this provider (e.g. iterate `Challenges` or maintain a per-provider pending challenge counter)
- Or keep the cooldown but make it shorter / configurable for dev chains

danielbui12:
@mudigal OMG, I've forgotten that part. As I already mentioned "deregister_provider ignores pending checkpoint rewards, unclaimed pool balance, and outstanding challenges" in https://github.com/paritytech/web3-storage/pull/80;

It seems that 2-step deregister is still the better choice; no need to add more storage interface / Challenges iterations.

danielbui12:
Closed with final veridiction https://github.com/paritytech/web3-storage/pull/172#issuecomment-4717774264

Provider need to wait for Challenges to expire before completely leaving
```
        /// Process expired challenges at the end of each block.
        fn on_finalize(n: BlockNumberFor<T>) {
            // Check if there are any challenges expiring at this block
            if let Some(expired_challenges) = Challenges::<T>::take(n) {
                for (index, challenge) in expired_challenges.iter().enumerate() {
                    // Slash the provider for failing to respond
                    let challenge_id = ChallengeId {
                        deadline: n,
                        index: index as u16,
                    };
                    Self::slash_provider_for_failed_challenge(challenge, challenge_id);
                }
            }
        }
```

PR #179, opened 2026-06-15
feat(pallet): challenge cost escalation (#157)

### LINKED ISSUE #157: Implement challenge cost escalation (cost ladder) for griefing prevention
## Summary

The challenge deposit is currently hardcoded to a fixed value of `100u32` in `pallet/src/lib.rs` (line ~3742), with the comment `"simplified - would be based on expected costs"`. The presentation (W3S-2026-Storage) describes a progressive cost escalation mechanism where challenges become more expensive for repeat challengers to prevent griefing — this is not yet implemented.

## Current Behavior

```rust
// Calculate deposit (simplified - would be based on expected costs)
let deposit: BalanceOf<T> = 100u32.into();
```

The `challenges_received` statistic is tracked on the provider but not used to adjust challenge costs.

## Expected Behavior

Challenge deposits should escalate based on factors such as:
- Number of previous challenges by the same challenger against the same provider/bucket
- Frequency of challenges within a time window
- Outcome of prior challenges (failed challenges should increase cost more)

This cost ladder is the core economic argument against griefing in the challenge game — without it, an attacker can spam cheap challenges to harass honest providers.

## References

- Pallet source: `pallet/src/lib.rs`, `create_challenge()` function

mudigal:
## Summary

- Replace hardcoded challenge deposit (100) with per-(challenger, provider) exponential escalation: `BaseChallengeDeposit × 2^min(active + failed, 8)`, capping at 256× base
- When provider defends, challenger's `failed` count increments (escalation persists)
- When provider is slashed, challenger's history is cleared (reset to base)
- Add `deposit` field to `ChallengeCreated` event

## Changes

- `BaseChallengeDeposit` config constant (100 units in mock, 100 UNIT in runtimes)
- `ChallengerRecord` struct + `ChallengeHistory` storage double-map
- Updated `create_challenge`, `respond_to_challenge`, `slash_provider_for_failed_challenge`
- 5 new escalation tests; all 184 pallet tests pass

## Test plan

- [x] `cargo test -p pallet-storage-provider` — 184/184 pass
- [x] `cargo clippy` on pallet + drive-registry + s3-registry + both runtimes — clean
- [x] `/format` — all checks pass

Closes #157


ilchu (commented):
Judging from description only, this is a solid change. But I don't see any mentions of such escalation policy in #231, so I believe we're leaving it fixed for now, and doing the auth / non-auth challenger split only? Or in a way, that split and escalation can be orthogonal. What would be our stance here? @mudigal @eskimor

eskimor:
I don't understand why we want to escalate the deposit? The deposit is meant to be large enough to cover the cost of the challenge response (with some leeway to cover cost fluctuations. 10x might be reasonable). Any excess is also refunded.

For the actual cost: I don't see the reason for any artificial inflation. We just charge the real cost, which will go up with chain load anyways. So if someone keeps sending challenges, the chain can either handle it just fine -> all good  - or chain fees go up -> also fine.

ilchu:
> I don't understand why we want to escalate the deposit? The deposit is meant to be large enough to cover the cost of the challenge response (with some leeway to cover cost fluctuations. 10x might be reasonable). Any excess is also refunded.
> 
> For the actual cost: I don't see the reason for any artificial inflation. We just charge the real cost, which will go up with chain load anyways. So if someone keeps sending challenges, the chain can either handle it just fine -> all good - or chain fees go up -> also fine.

Understood. Then I guess it's safe to abandon this one.

PR #194, opened 2026-06-16
Fix challenge history crash from stale localStorage

mudigal:
## Summary

- Challenges loaded from localStorage may have missing fields if the schema changed between versions
- Accessing undefined fields (e.g. `.toLocaleString()` on `undefined`) in the table renderer crashes the Challenges page silently
- Default all fields when deserializing from localStorage so cached data from older versions is safe

## Test plan

- [ ] Clear localStorage, load Challenges page — no crash, empty state shown
- [ ] With old-format challenge data in localStorage — page renders with defaults
- [ ] New challenges populate all fields correctly

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/user-interfaces/provider/src/state/provider.state.ts
+++ b/user-interfaces/provider/src/state/provider.state.ts
@@ -135,7 +135,19 @@ const CHALLENGES_GENESIS_KEY = 'provider-challenges-genesis'
 function loadPersistedChallenges(): Challenge[] {
   try {
     const raw = localStorage.getItem(CHALLENGES_STORAGE_KEY)
-    return raw ? JSON.parse(raw) : []
+    if (!raw) return []
+    const parsed = JSON.parse(raw) as Partial<Challenge>[]
+    return parsed.map((c) => ({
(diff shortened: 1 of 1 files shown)


ilchu (commented):
I'd rather steer away from having schema hand encoded via some rehydration. Or if we really want to land it, we should probably enforce the schema with versions via something like [zod](https://zod.dev/) so there's less chance to experience drift when anything challenge-related moves sideways and this snippet is forgotten.

And while we're on this, let's talk about the role of local storage in UI overall: if I remember correctly, it was introduced to keep challenges that have already expired visible to the provider in UI. But it should not become a source of truth that we rely on, only a cache in the best case. And what would make most sense is probably reconstructing the resolved challenges via querying some block range on chain for respective events, or just straight out abandoning this functionality and only leaving unresolved ones visible as it's ultimately the way it's meant to be.

ilchu:
If you don't mind, I'll close this one. There's a followup for further investigations here: #238.

PR #195, opened 2026-06-16
Separate start-provider from on-chain registration

mudigal:
## Problem

Currently `just start-provider` automatically registers the provider on-chain before starting the node. This has two issues:

1. **Product**: Registration is a significant on-chain decision — the operator chooses how much stake to lock, what capacity to declare, pricing, agreement duration bounds, etc. Bundling it into the start command means the provider has no say in these parameters; they're silently set to defaults. This isn't right from a product capability perspective, especially when providers need to run specific commands on their nodes or configure settings before going live.

2. **Development**: When testing the registration flow itself (e.g. via the provider dashboard UI or demos), having `start-provider` auto-register makes it impossible to start from an unregistered state. You can't demo or test the registration experience because the provider is already registered by the time the node is up.

## Context

The bundling was introduced in PR #69 (`6d213d3`, May 2026, @danielbui12) as a CI convenience fix. The `s3-demo-ci` and `fs-demo-ci` tests were only passing by accident — they relied on `just demo` having already run `full-flow.js`, which registers the provider as a side effect. When run standalone, they'd fail. Rather than fixing each CI test to register independently, registration was bundled into `start-provider` so any test starting a provider would automatically have it registered on-chain. It was a CI shortcut, not an intentional product decision.

## Solution

- `just start-provider` now only starts the provider HTTP node — no on-chain registration
- `just start-n-reg-provider` preserves the old combined behavior for convenience during dev, where you just want a working provider quickly without thinking about it

### Commands after this change

| Command | What it does |
|---------|-------------|
| `just start-provider` | Starts the provider HTTP node only |
| `just register-provider` | Registers on-chain (standalone, idempotent) |
| `just start-n-reg-provider` | Registers then starts (old combined behavior, for dev convenience) |

## Test plan

- [ ] `just start-provider` starts the node without calling `register_provider`
- [ ] `just start-n-reg-provider` registers then starts (same as old behavior)
- [ ] `just register-provider` works standalone

🤖 Generated with [Claude Code](https://claude.com/claude-code)


justfile
```diff
@@ -212,8 +209,13 @@ start-provider MODE="inmemory" PORT=PROVIDER_PORT STORAGE_PATH="./provider-data"
         --enable-checkpoint-coordinator \
         $EXTRA_ARGS
 
+# Register on-chain then start the provider node (original behavior)
+start-n-reg-provider MODE="inmemory" PORT=PROVIDER_PORT STORAGE_PATH="./provider-data" KEYFILE="":
```
danielbui12:
```suggestion
register-then-start-provider MODE="inmemory" PORT=PROVIDER_PORT STORAGE_PATH="./provider-data" KEYFILE="":
```

justfile
```diff
@@ -212,8 +209,13 @@ start-provider MODE="inmemory" PORT=PROVIDER_PORT STORAGE_PATH="./provider-data"
...
+    just start-provider MODE="{{MODE}}" PORT="{{PORT}}" STORAGE_PATH="{{STORAGE_PATH}}" KEYFILE="{{KEYFILE}}"
+    just register-provider "{{KEYFILE}}"
+
 # Register provider on-chain (idempotent). Requires a running chain.
-# Usually called automatically by start-provider.
+# Called automatically by start-n-reg-provider, or run standalone.
```
danielbui12:
```suggestion
# Called automatically by register-then-start-provider, or run standalone.
```

danielbui12 (approved):

bkontur (approved):

x3c41a (approved):

Issue #178, opened 2026-06-15
Repo structure: split provider-node, break the storage-client dependency, tidy crate layout


bkontur:
## Why
- `provider-node` depends on `storage-client` — an inverted edge (server pulling in the client SDK).
- `provider-node` is one ~8.4k-LOC crate mixing storage, HTTP, and on-chain coordinators.
- Crate layout is inconsistent (`runtime/` vs `runtimes/`, dir names != crate names, flat vs nested).

Also sets up a clean home for `utils/storage-cli` (#175).

## 1. Break `provider-node` -> `storage-client`
What it actually pulls from `storage-client` is two unrelated things:
- **Pure wire types** (no I/O): negotiate `agreement::{NegotiateRequest, AgreementTermsOf, SignedTerms}` + `sign_terms` + `discovery::ProviderInfo`, used server-side in the `/negotiate` handler.
- **A second chain client**: `chain_state_coordinator.rs` uses `storage_client::ProviderClient` to talk to the parachain over WS — `get_provider_info`, `fetch_replay_hsn`, `fetch_request_timeout`, plus a finalized-block subscription parsed with `StorageProviderEventParser` + `substrate::PALLET_NAME`.

⚠️ **The provider already has its own chain client.** `subxt_client.rs` (`SubxtChainClient`, ~917 LOC) talks to the same parachain directly via subxt. So today there are **two parallel chain-access paths** in one node — the node's own `SubxtChainClient` *and* the reads/subscription routed through `storage_client::ProviderClient`. Step 1 isn't just deleting a types dependency; it collapses those two clients into one.

- [ ] Extract `storage-negotiation` primitive crate (wire types + `MultiSignature` serde + `sign_terms`); `AgreementTerms` etc. already live in `primitives`.
- [ ] Move the chain reads + event subscription (`get_provider_info`, `fetch_replay_hsn`, `fetch_request_timeout`, finalized-block stream) onto the node's own `SubxtChainClient` (post-#159), retiring the `ProviderClient` path.
- [ ] Drop `storage-client` from `provider-node`.

## 2. Split `provider-node` -> `provider/{storage,rpc,node}`

| Crate | Modules | subxt? |
|---|---|---|
| `provider/storage` | `storage/*`, `fs_index`, `s3_index`, `mmr` | no |
| `provider/rpc` (HTTP API) | `api`, `fs_api`, `s3_api`, `auth`, `negotiate`, `error`, `types` | no |
[...]

Issue #176, opened 2026-06-15
Make the Rust coverage gate robust and diagnostic (filename-based exclusion + whole-crate ratio is fragile)


bkontur:
## Problem

The `Rust coverage` job (`.github/workflows/check.yml`) fails on **any** decrease in a
whole-crate coverage **ratio**, with the chain-client layer excluded by a **filename
suffix** (`--ignore-filename-regex='(…|_subxt\.rs|command\.rs|cli\.rs|main\.rs)'`) and
reported as a single `--summary-only` number. This makes the gate fragile and
low-signal:

1. **Pure refactors move the number.** PR #159 (reuse one `SubxtChainClient`) consolidates
   `challenge_responder_subxt.rs` / `checkpoint_coordinator_subxt.rs` /
   `replica_sync_coordinator_subxt.rs` (all excluded by `_subxt\.rs`) into
   `subxt_client.rs` (~915 LOC), which the regex no longer matches. Those ~0%-covered
   lines now count, so provider coverage drops **63.88% → 59.08%** with zero behavior
   change and the gate fails. The exclusion is keyed on a filename suffix, so any
   rename/split/merge silently changes the baseline.

2. **The denominator is full of code that can't be unit-tested without a live chain.**
   On `dev` today `checkpoint_coordinator.rs`, `replica_sync_coordinator.rs`,
   `challenge_responder.rs`, and `replica_sync.rs` are already at **0%** — they only run
   against a real chain, so they're dead weight in the ratio.

3. **`--summary-only` hides the cause.** CI prints one aggregate `%`; "coverage decreased"
   gives no hint which file moved.

### Reproduced locally
`cargo llvm-cov` with the CI flags reproduces `dev` provider coverage = **63.88%**. The
per-file table shows the culprits clearly (coordinator files at 0%); the `*_subxt.rs`
files are excluded only by the regex. Including them (i.e. what `subxt_client.rs` does)
drags the total to ~59%.

## Proposed improvements

**Make the gate diagnostic**
- [ ] Drop `--summary-only`; write the full per-file coverage table to
      `$GITHUB_STEP_SUMMARY` (and/or a sticky PR comment) so a drop names the exact file.
- [ ] Upload `cargo llvm-cov --html` as a CI artifact for line-level drill-down.

**Measure the right thing**
- [ ] Gate on **patch/diff coverage** (coverage of lines changed in the PR) instead of the
      whole-crate ratio — immune to renames/refactors, and it points at untested *new*
[...]

PR #196, opened 2026-06-16
ci(coverage): gate on patch coverage, make the job diagnostic and robust

bkontur:
Reworks the Rust coverage gate per #176. The old gate compared a whole-crate coverage **ratio** (PR vs base), excluded the chain layer by **filename suffix** (`_subxt\.rs`), and printed a single `--summary-only` number — so pure refactors/renames moved the number (e.g. PR #159: 63.88% → 59.08% with zero behaviour change) and failures named no file.

## What changed (`.github/workflows/check.yml`)

**Measure the right thing — patch/diff coverage**
- Gate on **`diff-cover`** over the lines this PR changes. Renames/refactors of already-covered (or excluded) code no longer move the number, and the report names the exact uncovered lines.

**Diagnostic**
- Per-file coverage table + the `diff-cover` report written to the **job summary**.
- `lcov.info` + an **HTML report** uploaded as artifacts for line-level drill-down.

**Robustness / best practices**
- One instrumented test run rendered to lcov/html/text (no re-running tests per format).
- `lcov` source paths rewritten repo-relative so `diff-cover` matches git paths.
- Dropped the base-branch `checkout`/`stash` dance and the brittle `grep '^TOTAL' | awk` parsing.
- `set -euo pipefail` + explicit hard failures — no silent zero-coverage fallthrough; pinned `diff-cover`.
- Exclusions anchored by path/name and documented; full path/module-based exclusion tracked in #178.

`MIN_PATCH_COV` (min % of changed lines covered) starts at **80** and is a single tunable env var.

## Validation
- YAML parse + `actionlint` clean.
- End-to-end locally: generated `lcov`, confirmed the path-strip is required/correct, and ran `diff-cover` against an older ref — it matched paths and produced the per-file + missing-line report with all flags. (diff-cover 7.7.0 locally; CI uses 9.x, a superset.)

Addresses #176.


.github/workflows/check.yml
```diff
@@ -179,127 +179,106 @@ jobs:
...
+          echo "$HOME/.local/bin" >> "$GITHUB_PATH"
 
-      - name: Measure base branch coverage
+      - name: Measure coverage
         env:
-          BASE_SHA: ${{ github.event.pull_request.base.sha }}
+          # Coverage exclusions, anchored by path/name: generated code, pure-type
+          # primitives, and the chain-access layer that can only run against a live
+          # chain. Migrating these to a path/module-based exclusion is tracked in #178.
+          COV_IGNORE: '(/\.cargo/|/rustc/|weights\.rs|runtime_api\.rs|mock\.rs|benchmarking\.rs|bechmarking\.rs|/primitives/|subxt_client\.rs|_subxt\.rs|checkpoint_coordinator\.rs|replica_sync_coordinator\.rs|replica_sync\.rs|challenge_responder\.rs|/command\.rs|/cli\.rs|/main\.rs)'
```
mudigal: Why are we ignoring all coordinators here?

.github/workflows/check.yml
```diff
@@ -180,126 +183,102 @@ jobs:
...
+          cargo llvm-cov --no-report \
             -p pallet-storage-provider \
             -p pallet-drive-registry \
             -p pallet-s3-registry \
-            --ignore-filename-regex='(weights\.rs|runtime_api\.rs|mock\.rs|primitives/|benchmarking\.rs|bechmarking\.rs)' \
-            --summary-only 2>&1 \
-            | grep '^TOTAL' | awk '{print $10}' | tr -d '%')
-          if [ -z "$BASE_PALLET_COV" ]; then
-            echo "ERROR: Failed to measure base pallet coverage (build or test failure?)"
+            -p storage-provider-node
```
mudigal: One concern on what the coverage measures:
- Coverage should target exposed APIs, not internal helpers

The old code was deliberate about this - it collected coverage from tests that enter through the public surface:
 - Pallets: `cargo llvm-cov --lib` — runs lib tests that call through dispatchable extrinsics (`Pallet::register_provider()`, `Pallet::create_bucket()`, etc.)
 - Provider node: explicitly targeted `--test *_integration` - these spin up real HTTP servers and hit endpoints (`/upload`, `/download`, `/health`) via `reqwest`

The new code replaces both with a single `cargo llvm-cov --no-report -p storage-provider-node` which runs all test targets indiscriminately. This loses the intentional separation - coverage now includes unit tests that may call internal helpers directly, which can be misleading. You can hit 100% on an internal function but never test the extrinsic or HTTP endpoint that calls it, leaving the actual contract unverified.

Collecting coverage from true integration tests (zombienet) isn't practical — the binaries run as separate processes and Rust's instrumentation doesn't cross that boundary. That's fine. But the tests we do collect from should still enter through the public surface:
  - Pallet extrinsics — the dispatchable calls, not internal `do_*` helpers
  - Provider HTTP endpoints — tested with a faked chain client and storage backend (which is exactly what the `*_integration.rs` tests already do)
  - Client SDK public methods — what consumers actually call

These all run in cargo test with mocked/faked dependencies. The key difference is where the test enters.

Since this PR is reworking the coverage gate, it should preserve the existing behavior of measuring coverage at the API entry-point level. The gate should answer: "are the new lines reachable from a real API call?" rather than "did some test touch this line?"

.github/workflows/check.yml
```diff
@@ -180,126 +183,102 @@ jobs:
...
+            lcov.info
+            target/llvm-cov-html
+          if-no-files-found: warn
 
-      - name: Compare coverage
+      - name: Patch coverage gate
+        env:
+          BASE_REF: ${{ github.event.pull_request.base.ref }}
+          # Minimum coverage of the lines this PR changes. Tune as the suite grows.
+          MIN_PATCH_COV: '80'
```
mudigal, line 282: 80% patch coverage is a reasonable threshold. For context: lines that are genuinely untestable (e.g., error paths that require a live chain) should already be excluded via COV_IGNORE. If the team finds 80% too strict for certain PR types (e.g., pure-pallet PRs where mock coverage is straightforward vs provider-node PRs with harder-to-test async code), the single env var makes it easy to tune. Good design.

We should observe this and see if we need to change it after certain PRs.

.github/workflows/check.yml
ilchu: Because like the e.g. subxt client code they're untestable without a live chain.

.github/workflows/check.yml
ilchu: I'd say it's less fuss and maybe even more correct to cover everything, especially with the switch to diff coverage.

With targeted exclusion of not only purely untestable but also non-outside facing lines we impose a bigger burden of keeping the exclusion list up to date, which will eventually drift anyways and also don't stimulate ourselves to write proper unit tests even if the code is internal.

And re: the "integration" test take: coverage would not be able to push us to verifying the contract in any case, as it tests only the reach, but doesn't assert anything. For the latter we could eventually introduce prop tests or the like. And most crucially, for diff coverage the numerator has to be wide. If there is a PR where any line is covered only by a unit test, then it will turn into a false "uncovered line" failure in the CI. So if we're setting some cov target, then it has to be all the repo code and any kinds of tests with this change to diff-based. For contract adherence there's nothing but discipline and separate audits.

ilchu (approved):

.github/workflows/check.yml
mudigal: The subxt-layer files (subxt_client.rs, _subxt.rs) I agree with - those talk to the chain directly and are hard to test without one.

But the coordinators aren't in that bucket. checkpoint_coordinator, replica_sync_coordinator, and challenge_responder each abstract chain access behind a trait (CheckpointChainClient / ReplicaSyncChainClient / ChallengeChainClient) precisely so they can run against a mock, and we already have those tests in tests/coordinators/ (MockCheckpointChainClient et al. are plain in-memory structs, no chain). So excluding them from coverage on "needs a live chain" grounds isn't right - they're exactly the code that benefits from the gate.

replica_sync.rs is the one edge case: it has no trait of its own and is only exercised indirectly through the coordinator, so I'm less fussed about it - but that's a "not directly covered yet" reason, not a "needs a live chain" one.

Proposal: keep the subxt files excluded, drop the three coordinators from the ignore list (they're already tested), and leave replica_sync.rs out for now with a note that it should be covered once it's testable directly.

bkontur:
```
just coverage
```


Added script now generates something like this:

 | Module | Line coverage | Lines |                                                                                                                                                                                      
  |---|---|---|                                                                                                                                                                                                           
  | `pallet` | 88.48% | 2058/2326 |                                                                                                                                                                                       
  | `provider-node` | 86.65% | 4601/5310 |                                                                                                                                                                                
  | `provider/negotiation` | 100.00% | 37/37 |                                                                                                                                                                            
  | `storage-interfaces/file-system/pallet-registry` | 84.30% | 102/121 |                                                                                                                                                 
  | `storage-interfaces/s3/pallet-s3-registry` | 67.33% | 136/202 |                                                                                                                                                       
  | **total** | **86.72%** | 6934/7996 |                                                                                                                                                                                  
                                                                                                                                                                                                                          
  Integration-only coverage (public-API view):                                                                                                                                                                            
                                                                                                                                                                                                                          
  Line coverage reachable through integration tests alone —                                                                                                                                                               
  diagnostic only, no gate.                                                                                                                                                                                               
                                                                                                                                                                                                                          
  | Module | Line coverage | Lines |                                                                                                                                                                                      
  |---|---|---|                                                                                                                                                                                                           
  | `pallet` | 0.00% | 0/2326 |                                                                                                                                                                                           
  | `provider-node` | 64.03% | 3400/5310 |                                                                                                                                                                                
  | `provider/negotiation` | 43.24% | 16/37 |                                                                                                                                                                             
  | `storage-interfaces/file-system/pallet-registry` | 0.00% | 0/121 |                                                                                                                                                    
  | `storage-interfaces/s3/pallet-s3-registry` | 0.00% | 0/202 |                                                                                                                                                          
  | **total** | **42.72%** | 3416/7996 |                                                                                                                                                                                  
  -------------

bkontur:
@mudigal @ilchu closing out the collection-scope question:                                                                                                                                                            
                                                                                                                                                                                                                          
The open question was whether the coverage gate should count only integration tests (proves the public API is exercised, but fails legitimately unit-tested lines — every pallet line, for instance, since pallets are  tested through mock runtimes, not HTTP) or all tests (no false failures, but a PR can pass without touching the endpoint). Neither is perfect, so rather than a half-measure the gate stays strict and simple: all tests, full diff — which matches Ilia's take on the thread.                                                                                                                                                             
                                                                                                                                                                                                                          
  For the public-API question, this PR also adds a dedicated integration-only view: the same test run rendered a second time from the integration-test profiles alone — no test runs twice, no second build. It shows per module what is reachable purely through the public surface, next to the merged table in the job summary, plus an lcov-integration.info artifact. Diagnostic only, not gated: it's a strict subset of the merged profile, so it can never change a gate outcome, and pallets legitimately sit at 0% there. First run already earned its keep — current dev baseline:                                                              
  
| Module | All tests | Integration only |                                                                                                                                                                               
  |---|---|---|                                                                                                                                                                                                           
  | `pallet` | 88.48% (2058/2326) | 0.00% (0/2326) |                                                                                                                                                                      
  | `provider-node` | 86.65% (4601/5310) | 64.03% (3400/5310) |                                                                                                                                                           
  | `provider/negotiation` | 100.00% (37/37) | 43.24% (16/37) |                                                                                                                                                           
  | `storage-interfaces/file-system/pallet-registry` | 84.30% (102/121) | 0.00% (0/121) |                                                                                                                                 
  | `storage-interfaces/s3/pallet-s3-registry` | 67.33% (136/202) | 0.00% (0/202) |                                                                                                                                       
  | **total** | **86.72%** (6934/7996) | **42.72%** (3416/7996) |                                                                                                                                                                                                                               

 **One caveat on that view**: our tests/ suites build the axum router in-process with mocked chain access — they don't run the real provider-node binary. So "reachable through the public API" is still an approximation; a fully honest version would drive the actual binary over HTTP (the way just demo does), and that's out of reach for in-process coverage instrumentation without extra machinery. Worth keeping in mind before anyone  reads too much into the second column — it's tricky either way (dobale - but it needs graceful-shutdown handling (SIGKILL loses the profile data) and process orchestration in CI,).
                                                                                                                                                    
                                                                                                                                                                                                                          
  **My honest take:** I'm still not convinced about integration-tests-only gating for the provider, and I'm not fully sure about the benefit of the coverage gate itself — coverage is about reach; it proves code ran, not that it's correct. I think we should put more weight on correctness and integration tests — that's the direction of #265 (Quint spec for the challenge protocol, checking invariants like honest-provider-never-slashed, which no coverage number can). The two can work together though, so let's go with the diff gate for now and evolve from there.

mudigal:
> **My honest take:** I'm still not convinced about integration-tests-only gating for the provider, and I'm not fully sure about the benefit of the coverage gate itself — coverage is about reach; it proves code ran, not that it's correct. I think we should put more weight on correctness and integration tests — that's the direction of #265 (Quint spec for the challenge protocol, checking invariants like honest-provider-never-slashed, which no coverage number can). The two can work together though, so let's go with the diff gate for now and evolve from there.

Thanks @bkontur — agree on landing the diff gate now and evolving from there. Let me put down what I think the gate is for, because I think it reframes the "all tests vs integration-only" question.

The value isn't "a test touched this line" — it's "this line is reachable through one of the system's exposed interfaces." That's the property that protects us: it proves the code is wired into the product and not just into a test, so it catches dead code that nothing in the real call graph reaches. That's the failure mode I most worry about now that a lot of code is generated — a private helper gets a direct unit test, shows as covered, then a refactor drops its only public caller. The code is dead, but "all tests, full diff" still counts it green; an interface-driven view flags it.

The distinction I care about isn't HTTP, though — it's "exposed interface," which is per-component. For the provider-node that's the axum router; for a pallet it's the extrinsics called through the mock runtime, which is the pallet's public surface. So I'm not arguing for HTTP-only gating (that's what unfairly zeroes out the pallets in the integration-only column) — just that coverage should come from entering through each component's own public interface.

And mocks were never the issue. I'm fine with mocked chain access and the in-process router — the router is the exposed interface, so driving it in-process with a mocked chain satisfies the goal. We don't need the real binary over HTTP in CI to get the property we want; the requirement is "entered via the public surface," not "run against a live system."

For the code that genuinely can't run without a network/chain call — the subxt layer that instantiates the real connection — I'm all for excluding those specific files, kept to a minimal reviewed list. That's different from the coordinators earlier in the thread, which abstract the chain behind a trait and are already mock-tested through their public interface; those should stay in.

And fully agree coverage proves reach, not correctness — #265 is where the real assurance comes from. I see them as complementary: the gate is a hygiene floor (no dead code, every path reachable through the public surface), the spec work is the correctness proof. Happy to run the diff gate as-is now and move toward interface-driven measurement plus the exclusion list as a follow-up.

scripts/coverage.sh
```diff
@@ -0,0 +1,205 @@
...
+	cargo llvm-cov report --html --output-dir target/llvm-cov-html --ignore-filename-regex "$COV_IGNORE"
+	cargo llvm-cov report --ignore-filename-regex "$COV_IGNORE" >coverage-summary.txt
+
+	if [ ! -s lcov.info ] || [ ! -s lcov-integration.info ]; then
+		echo "error: lcov.info or lcov-integration.info is empty — coverage generation failed" >&2
+		exit 1
+	fi
+
+	# Make lcov source paths repo-relative so diff-cover matches git paths.
+	sed -i "s|SF:${REPO_ROOT}/|SF:|g" lcov.info lcov-integration.info
```
mudigal: `sed -i` with no suffix arg is GNU-only. On BSD/macOS `sed` this consumes the script as a backup suffix and fails - so just coverage (advertised below as reproducing the CI gate locally) breaks at this step on the macs this team develops on. Portable form:

```suggestion
     # Make lcov source paths repo-relative so diff-cover matches git paths.
     sed -i.bak "s|SF:${REPO_ROOT}/|SF:|g" lcov.info lcov-integration.info
     rm -f lcov.info.bak lcov-integration.info.bak
``` 

Can you confirm a local just coverage run passes on macOS before merge?

scripts/coverage.sh
```diff
@@ -0,0 +1,205 @@
...
+	echo
+	cat coverage-modules.md
+	echo
+	echo "Integration-only coverage (public-API view):"
+	echo
+	cat coverage-integration.md
+}
+
+gate() {
+	need diff-cover "python3 -m pip install 'diff-cover>=9,<10'"
```
mudigal: CI pins `diff-cover >=9, <10`, but just coverage only checks presence - a contributor on 7.x (the version the rest of this PR was validated against) gets --format markdown: behavior that differs from CI. Worth a lightweight version assertion in need, or at least documenting the required major version so local and CI can't silently diverge.

scripts/coverage.sh
```diff
@@ -0,0 +1,205 @@
...
+	# subset, so run integration targets first, render the integration-only
+	# view, then run the remaining targets (--lib --bins covers exactly
+	# what the single default-target invocation ran) and render the merged
+	# view. The upfront clean keeps a local rerun's stale profraws out of
+	# the first view. A build/test failure aborts here — no silent
+	# zero-coverage fallthrough.
+	cargo llvm-cov clean --profraw-only
+	# --test '*' = every target under tests/ — picks up new suites
+	# automatically, skips packages that have none.
+	cargo llvm-cov --no-report "${pkg_flags[@]}" --test '*'
```
mudigal, line 137: The comment says this "skips packages that have none," which holds per-package - but with all -p flags combined, if no test target across the whole selected set matched the glob, cargo errors with "no test target matches pattern." Non-empty today (provider-node), so fine now; it's a latent break if the measured set ever shrinks. Does provider-negotiation actually expose a tests/ target under this glob? A one-line note on the invariant (or || true guard) would harden it.

justfile
```diff
@@ -60,6 +60,13 @@ build-paseo-runtime:
 build-provider:
     cargo build --release -p storage-provider-node
 
+# Reproduce the CI Rust coverage gate locally (needs cargo-llvm-cov + diff-cover)
+coverage:
+    #!/usr/bin/env bash
+    set -euo pipefail
+    git fetch --no-tags origin dev
```
mudigal, line 68: Correct for the usual base, and COMPARE_BRANCH overrides the script - but the recipe as written can't reproduce a PR whose base isn't dev. Fine to leave; maybe note the COMPARE_BRANCH=… override in the recipe comment for the non-dev case.

scripts/coverage.sh
bkontur: > Can you confirm a local just coverage run passes on macOS before merge?

@danielbui12 thank you both, MacOS works also now

justfile
bkontur: > Correct for the usual base, and COMPARE_BRANCH overrides the script - but the recipe as written can't reproduce a PR whose base isn't dev. Fine to leave; maybe note the COMPARE_BRANCH=… override in the recipe comment for the non-dev case.

thank you, ok, added: https://github.com/paritytech/web3-storage/pull/196/changes/f1eb49c3b3e459aa4239bad27465439aae6691f4

scripts/coverage.sh
bkontur: > CI pins `diff-cover >=9, <10`, but just coverage only checks presence - a contributor on 7.x (the version the rest of this PR was validated against) gets --format markdown: behavior that differs from CI. Worth a lightweight version assertion in need, or at least documenting the required major version so local and CI can't silently diverge.

thank you, fixed https://github.com/paritytech/web3-storage/commit/f1eb49c3b3e459aa4239bad27465439aae6691f4 - shared minimal version is used now from config with some gate, should not diverge now

PR #197, opened 2026-06-16
Skip CI for docs-only changes

mudigal:
## Problem

All CI workflows run on every push/PR regardless of what changed. A README or docs update triggers the full Rust build, integration tests, and runtime migration checks — wasting CI minutes and blocking PRs unnecessarily.

## Solution

Add `dorny/paths-filter` change detection to the three heaviest workflows, using the same pattern already established in `ui-checks.yml`:

| Workflow | Skips when only these change | Always runs on |
|---|---|---|
| `check.yml` | Non-Rust files (docs, markdown, .claude/) | `workflow_dispatch` |
| `integration-tests.yml` | `*.md`, `docs/**`, `.claude/**`, `LICENSE` | `workflow_dispatch` |
| `check-runtime-migration.yml` | Non-runtime/pallet files | `workflow_dispatch`, `schedule` |

### How it works

1. A lightweight `changes` job runs `dorny/paths-filter` to detect what changed
2. Expensive jobs (build, test, clippy, migrations) get `if: needs.changes.outputs.<filter> == 'true'`
3. Downstream jobs auto-skip via dependency chain (e.g. clippy/test/benchmarks depend on `check`)
4. Aggregate gate jobs (`basic-checks`, `Integration Tests`, `All migrations passed`) always report with `if: always()` — so branch protection stays safe even when everything is skipped

### Not changed

- `zizmor.yml` — already lightweight (ubuntu-latest, just a CLI tool)
- `ui-checks.yml` — already has path filtering
- `deploy-ui.yml` — already has top-level `paths` filter

## Test plan

- [ ] A docs-only PR skips Rust checks and integration tests
- [ ] A Rust code PR runs all checks normally
- [ ] `workflow_dispatch` always runs everything
- [ ] Gate jobs always report (branch protection not broken)
- [ ] Scheduled runtime migration snapshots still refresh daily

### DIFF
--- a/.github/workflows/check-runtime-migration.yml
+++ b/.github/workflows/check-runtime-migration.yml
@@ -48,6 +48,7 @@ jobs:
           filters: |
             runtime:
               - 'runtime/**'
+              - 'runtimes/**'
               - 'pallet/**'
               - 'primitives/**'
               - 'storage-interfaces/**/pallet-*/**'
--- a/.github/workflows/check-runtime-migration.yml
+++ b/.github/workflows/check-runtime-migration.yml
@@ -48,6 +48,7 @@ jobs:
           filters: |
             runtime:
               - 'runtime/**'
+              - 'runtimes/**'
               - 'pallet/**'
               - 'primitives/**'
               - 'storage-interfaces/**/pallet-*/**'
@@ -57,6 +58,8 @@ jobs:
 
   runtime-matrix:
     runs-on: ubuntu-latest
+    needs: [changes]
+    if: needs.changes.outputs.runtime == 'true'
     permissions:
       contents: read
     outputs:
--- a/.github/workflows/check-runtime-migration.yml
+++ b/.github/workflows/check-runtime-migration.yml
@@ -29,6 +29,32 @@ jobs:
       contents: read
     uses: ./.github/workflows/set-image.yml
 
+  # Skip migration checks when only docs/non-runtime files changed.
+  # Always runs on schedule (snapshot refresh) and workflow_dispatch.
+  changes:
+    name: Detect changes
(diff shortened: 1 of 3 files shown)


danielbui12 (approved):

.github/workflows/check-runtime-migration.yml
```diff
@@ -29,6 +29,32 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+        with:
+          persist-credentials: false
+      - uses: dorny/paths-filter@fbd0ab8f3e69293af611ebaee6363fc25e6d187d # v3
+        id: filter
+        with:
+          filters: |
+            runtime:
+              - 'runtime/**'
```
bkontur, line 50:
```suggestion
              - 'runtime/**'
              - 'runtimes/**'
```

.github/workflows/check-runtime-migration.yml
```diff
@@ -159,10 +186,10 @@ jobs:
 
   check-migrations:
     name: check-migration (${{ matrix.runtime.name }})
-    needs: [set-image, runtime-matrix, prepare-snapshots]
+    needs: [set-image, runtime-matrix, prepare-snapshots, changes]
     runs-on: parity-large
     timeout-minutes: 120
-    if: github.event_name != 'schedule' && github.event_name != 'workflow_dispatch'
+    if: github.event_name != 'schedule' && github.event_name != 'workflow_dispatch' && needs.changes.outputs.runtime == 'true'
```
bkontur, line 194: We could/should put this múch sooner ayound line 59
 Why runtime-matrix step starts

PR #207, opened 2026-06-17
FollowUp: Move agreement negotiation off-chain, keep only mutual commitment on-chain

### LINKED ISSUE #124: Provider-side negotiation workflow (manual approve / reject / counter)
**Follow up of https://github.com/paritytech/web3-storage/issues/97**

## Context

Storage agreements are established via a two-step flow:
1. the bucket/drive admin POST a `NegotiateRequest` to the provider node `/negotiate` endpoint
2. the provider node returns `SignedTerms`, and the owner redeem them on chain via `establish_storage_agreement` (or the wrapping `create_drive` / `create_s3_bucket`).

Providers should able to negotiate through the **provider dashboard**: see the queue of pending requests and approve / reject / counter each one. I.e:
- review high-value requests before signing,
- offer a counter (e.g. higher price for an oversized capacity),
- reject specific owners (without flipping `accepting_primary` off for everyone),
- have an audit trail of who asked for what.

## Current State

- `POST /negotiate` (provider-node) auto signs and returns `SignedTerms` synchronously. The handler lives in `provider-node/src/api.rs` and uses the checkpoint key + nonce counter directly.
- Pending state isn't persisted; No queue, no history.
- Token bucket with IP as key rate limiter

## Proposed Approach

**Provider-node**
- POST `/negotiate` → Persist pending requests in the existing storage backend (disk mode / in-memory); Return `202 Accepted { request_id, status: "pending" }`.
- New provider-only endpoints (auth via the existing `Web3Storage <pubkey>:<sig>:<ts>` header used by other dashboard endpoints):
  - `GET /negotiations?status=pending|approved|rejected|expired`
  - `PATCH /negotiations/:id/approve` → issues a unique provider's nonce counter, marks `approved`
  - `PATCH /negotiations/:id/reject { reason? }`
- New owner-facing endpoint `GET /negotiations/:request_id` → polled by the owner SDK; returns `{ status, signed_terms?, nonce_counter? }`.
- Background expiry job: requests older than `RequestTimeout` flip to `expired`.

**New flow** (open to alternatives)
1. Bucket admin sends `NegotiateRequest`.
2. Provider goes to dashboard and able to review all agreement requests
3. Provider updates agreement request via dashboard. I.e provider approves an bucket agreement → status becomes `Approved` & provider node issue an unique nonce counter.
4. On accept → Bucket admin sends GET with `request_id` → provider auto-signs the previously-recorded terms & counter nonce. / Or when the Provider accept the agreements, then it is auto submitted via unsigned feeless transaction. (in the agreement term already includes the owner of bucket, ... necessary information)
5. Bucket admin redeems sig & terms on-chain to create bucket/drive.

### runtime
- [ ] Verify ProviderReplayState:
   - [ ] Is `pub const REPLAY_WINDOW_BITS: u32 = 256;` too short? when the Provider accept the agreements, if we implement auto submit unsigned feeless transaction (in the agreement term already includes the owner of bucket, ... necessary information) this won't be problematic.

### provider-node
- [ ] Define `NegotiationRecord` storage schema; (disk / in-memory) implementations.
- [ ] Implement new provider HTTP endpoints
   - [ ] `POST /negotiate`
   - [ ] `GET /negotiations?status=pending|approved|rejected|expired`
   - [ ] `PATCH /negotiations/:id/approve`
   - [ ] `PATCH /negotiations/:id/reject { reason? }`
- [ ] Implement new bucket admin HTTP endpoints:
   - [ ] `GET /negotiations/:request_id`
- [ ] Unit tests, Integration test, ...

### client SDK (`client/`)
- [ ] ...

### Console-ui / drive-ui
- [ ] ...

### provider-ui dashboard
- [ ] ...

danielbui12:
##  Primary Changes

On-chain validity-window enforcement
  - `valid_until` must be in range  [current block, current block + RequestTimeout]
  - Added an integrity_test asserting `RequestTimeout < DeregisterAnnouncementPeriod` — this is the core invariant that closes the re-register replay window (an old quote always expires before a provider can deregister + reregister).
  - `ProviderMultiaddrUpdated` event emit `multiaddr`.
  
Provider node: live chain-state sync
  - New `chain_state_coordinator.rs` introduces ChainState (atomic current_block + RwLock<provider_info>) holds state that frequently change. 
  - Fetch `RequestTimeout` from pallet constant at startup.
  - Resolve `valid_until = u32::MAX` in /negotiate; stamps valid_until = current_block + request_timeout
  - Migrate @bkontur's reconciler to chain_state_coordiantor

## Issue:

- Close https://github.com/paritytech/web3-storage/issues/124


danielbui12:
@mudigal there is a limitation in provider-node coverage: it is not able to run integration tests related to runtime due to the lack of blockchain startup in the check.yml job. Should we add start-chain to check.yml to increase test coverage, or do you want to keep it as simple as it is?

bkontur:
@danielbui12 my assistant says:
> The async connect loops are hard to unit-test, but parse_pallet_events and the relevance-filter in process_provider_events are pure and testable with hand-constructed events. Adding those should recover the gate.

pallet/src/lib.rs
```diff
@@ -82,6 +82,19 @@ pub mod pallet {
...
+
+        fn integrity_test() {
+            // The re-register replay defense relies on RequestTimeout being strictly
+            // shorter than DeregisterAnnouncementPeriod: a quote signed at block S
+            // expires at S+RequestTimeout, which is before the provider can complete
+            // deregistration and re-register (requiring DeregisterAnnouncementPeriod
+            // more blocks), so an old quote cannot be replayed against the new
+            // incarnation.
+            assert!(
+                T::RequestTimeout::get() < T::DeregisterAnnouncementPeriod::get(),
```
bkontur, line 94: this is just compile-time check, but we allow to change those with `pub storage` and `setStorage` root calls (those calls also does not catch this invariant).

We should definitely add `fn try_state` checks also, to hit the stuff when doing runtime upgrade at least (@danielbui12 please, create follow-up issue for adding try_state checks and cover some important).

examples/papi/full-flow.js
```diff
@@ -69,7 +69,7 @@ async function setupAgreement(api, providerUrl, client, provider) {
     owner: client.address,
     max_bytes: maxBytes,
     duration,
-    price_per_byte: 1n,
+    price_per_byte: 100n,
```
bkontur: @danielbui12 why 1->100?

client/src/provider.rs
```diff
@@ -209,6 +209,14 @@ impl ProviderClient {
                 .ok_or_else(|| ClientError::Chain("Missing 'accepting_extensions'".to_string()))?,
             agreements_total,
             challenges_failed,
+            deregister_at: named_field(&value, "deregister_at").and_then(|v| match &v.value {
```
bkontur, line 212: @danielbui12 @ilchu not sure, if this storage item parsing is the right thing, we should use runtime api's or something typed as mentioned here: https://github.com/paritytech/web3-storage/pull/125#discussion_r3442998786

client/src/provider.rs
```diff
@@ -844,6 +869,21 @@ pub struct ProviderSettings {
     pub max_capacity: u64,
 }
 
+/// Mirror StorageProvider::ProviderSettings
```
bkontur: @danielbui12 where do we need there `Default` impl? this can easily go out of sync. Can we remove this default impl?

bkontur:
@danielbui12 also please check the up-to-date docs, my assistant says:
> /negotiate gate weakened to is_some(); stale is_bootstrapped comments + misleading test comment — restore gate or fix comments

client/src/provider.rs
danielbui12: @bkontur `get_provider_info` read directly from Providers storage interface, so `deregister_at` definitely included there;
As my experience work in provider-node, a lot of subxt parsings are either out-of-date or error. I think we can have a separate PR to fix them all & create typed api or something else relavant.

pallet/src/lib.rs
danielbui12: Created follow-up issue https://github.com/paritytech/web3-storage/issues/230

danielbui12:
@ilchu 
Re-open: https://github.com/paritytech/web3-storage/pull/105#discussion_r3342108431
> I don't see how we would need a single file with the current nonce on disk. From what I understand, it would only generate unnecessary overhead while also allowing new failure modes -- mostly due to it relying on the same atomic counter as the backing. So the state could get desynced if e.g. the counter gets updated -> the provider crashes -> the file still has the old value.

Yes we do need to persis nonce, even persist agreement terms as well for several reasons:
- provider signed nonce & term, but that signature cannot be used because chain is unreachable, user need to try later within `valid_until`.
- provider signed nonce & term, but that signature is not submitted; then provider node crash due to unexpected behavior; when re-bootstrap, provider node need to be start from the latest signed nonce, not the one in runtime. If provider node re-start from the nonce in runtime, a single nonce may be used twice, whoever submits second is rejected as `AlreadyUsed`.

From my perspective, persisting nonce does not introduce more failure points. Because the nonce will following the database / persistence mode as provider selected.

client/src/provider.rs
bkontur: > As my experience work in provider-node, a lot of subxt parsings are either out-of-date or error. I think we can have a separate PR to fix them all & create typed api or something else relavant.

Yes, exactly, please @danielbui12  create an issue (if not exists) and let's find some solution with @ilchu, which should also have some test to ensure this, and also in the near future we will have some "compliance/conformance" tests for provider protocol impl

bkontur (approved):
lgtm - we will anyway revisit provider events/pooling mechanism later: https://github.com/paritytech/web3-storage/issues/222 + also synced/cached provider state

PR #208, opened 2026-06-17
feat: add pallet-utility

danielbui12:
## Changes 

add pallet-utility enabling batch/batchAll transactions


runtimes/web3-storage-local/src/lib.rs
```diff
@@ -562,6 +569,10 @@ mod runtime {
     // Smart contracts (PolkaVM + EVM-compatible)
     #[runtime::pallet_index(60)]
     pub type Revive = pallet_revive;
+
+    // Handy utilities. Utility / Multisig / Proxy / Indices ...
+    #[runtime::pallet_index(70)]
```
bkontur: I would put this as 34 and move, but also not that important

```suggestion
    #[runtime::pallet_index(34)]
```

bkontur (approved):
@danielbui12 just out of curiosity, do you have any scenario in mind?

danielbui12:
@bkontur  

I'm working on this  followup https://github.com/paritytech/web3-storage/pull/105#pullrequestreview-4485199478
"**Provider-side request vetting + replay-window sizing / auto-submit queue** — already tracked in #124."

This will introduce a manual agreement approval step on the provider node. Providers can optionally enable automatic approval via `--enable-auto-accept-agreement [false|true]`.

Once the provider manual approves the agreement, a new worker can handle feeless & unsigned transaction for `create_s3_bucket` or `create_drive` extrinsics. We could also use `batchAll` to execute the related extrinsics atomically, simplifying the client-side flow.

As mentioned in #124, I proposed `create_s3_bucket`, `create_drive`, `establish_storage_agreement` could be converted to feeless transactions since the agreement terms already contain the necessary information (bucket owner, provider).

PR #210, opened 2026-06-17
paseo: make storage-provider params runtime-settable via setStorage (v0.4.0-paseo)

mudigal:
## Summary

Makes the storage-provider economic/timing parameters tunable on a live previewnet via `system.setStorage` (sudo/governance) **without a runtime upgrade**, and drops the unused fast-runtime scaffolding. Production timeout defaults are unchanged.

### 1. Paseo parameters are runtime-settable (`pub storage`)

The paseo `parameter_types!` block is now storage-backed (`pub storage` instead of `pub const`). Each value remains its compile-time default but is read from unhashed storage, so any of them can be overridden live with `system.setStorage` under sudo/governance — no wasm redeploy:

`MinProviderStake`, `ChallengeTimeout`, `SettlementTimeout`, `RequestTimeout`, `MinStakePerByte`, `DefaultCheckpointInterval`, `DefaultCheckpointGrace`, `CheckpointReward`, `CheckpointMissPenalty`, `DeregisterAnnouncementPeriod`.

**Defaults are unchanged** (production values: 48h/24h/6h challenge/settlement/request, etc.). On previewnet we shrink whatever we need at runtime via setStorage — e.g. short challenge/deregistration windows for testing — and revert just as easily. The local (testnet) runtime keeps plain `pub const`.

> Each `pub storage X` resolves its value from `twox_128(b":X:")` and also generates `X::key()` / `X::set()`. A setStorage call sets that key to the SCALE-encoded value.

### 2. Drop the (inert) fast-runtime scaffolding

`dev` carried fast-runtime scaffolding that nothing consumed — a second `build.rs` wasm (`fast_runtime_binary`), the `fast_runtime_binary` module in `lib.rs`, and a `fast-runtime = []` Cargo feature — but the timeout constants were never cfg-gated, so the second binary was identical to the default and CI never built with the feature. With runtime params now tunable via setStorage, this scaffolding is unnecessary; removed from both the local and paseo runtimes.

### 3. Release bump: v0.4.0-paseo

- `storage-provider-node` crate 0.3.0 → 0.4.0 (release-tag validation checks this)
- `storage-paseo-runtime` crate 0.3.0 → 0.4.0
- `RuntimeVersion.spec_version` 3_000 → 4_000
- `Cargo.lock` updated

No extrinsic encoding changes, so `transaction_version` stays at 2; the `spec_version` bump triggers on-chain migrations on upgrade.

## Tests

`storage-paseo-runtime`'s XCM integration tests built their `RuntimeCall`s (via `default_stake()` → `MinProviderStake::get()`) *outside* `execute_with`. With the parameter now storage-backed, that read needs an Externalities environment, so the four affected tests now construct the stake/call **inside** the `execute_with` closure. (No behaviour change — production reads already happen inside externalities.)

## PreviewNet deployment

Deploy the v0.4.0-paseo runtime as usual (production defaults). Then, on the live chain, issue `sudo.setStorage([...])` to set the desired previewnet values for any of the parameters above — and adjust again anytime without a runtime upgrade.

## Test plan

- [x] `storage-parachain-runtime` (local) + `storage-paseo-runtime` (paseo) build (release)
- [x] `cargo test -p storage-paseo-runtime --test tests` — 19/19 pass (XCM tests fixed)
- [x] `cargo +nightly fmt --all --check` clean
- [x] Integration/E2E unaffected — defaults match `dev`, which is green

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9677,7 +9677,7 @@ dependencies = [
 
 [[package]]
 name = "storage-paseo-runtime"
-version = "0.3.0"
+version = "0.4.0"
 dependencies = [
  "cumulus-pallet-aura-ext",
  "cumulus-pallet-parachain-system",
@@ -9760,7 +9760,7 @@ dependencies = [
 
 [[package]]
 name = "storage-provider-node"
-version = "0.3.0"
+version = "0.4.0"
 dependencies = [
  "async-trait",
  "axum",
--- a/justfile
+++ b/justfile
@@ -48,13 +48,13 @@ default:
 build:
     cargo build --release
 
-# Build only the runtime
+# Build only the runtime (fast-runtime shrinks timeouts for local testing)
 build-runtime:
-    cargo build --release -p storage-parachain-runtime
+    cargo build --release -p storage-parachain-runtime --features fast-runtime
 
-# Build only the paseo runtime
+# Build only the paseo runtime (fast-runtime shrinks timeouts for local testing)
 build-paseo-runtime:
-    cargo build --release -p storage-paseo-runtime
+    cargo build --release -p storage-paseo-runtime --features fast-runtime
 
 # Build only the provider node
 build-provider:
--- a/justfile
+++ b/justfile
@@ -48,13 +48,13 @@ default:
 build:
     cargo build --release
 
-# Build only the runtime (fast-runtime shrinks timeouts for local testing)
+# Build only the runtime
 build-runtime:
-    cargo build --release -p storage-parachain-runtime --features fast-runtime
+    cargo build --release -p storage-parachain-runtime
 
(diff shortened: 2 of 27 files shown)


mudigal:
@bkontur Can we update the runtime in the previewnet? To demo the registration flow we would have to de-register and the period of 4 days is a lot.

bkontur:
> @bkontur Can we update the runtime in the previewnet? To demo the registration flow we would have to de-register and the period of 4 days is a lot.

@mudigal fast-runtime is pretty complication also for PreviewNet deployment, both runtimes are just testnet runtimes, so I set those values as you needed and made them `pub storage` so we can change them now with sudo setStorage calls - more flexibility for tesrting

bkontur (approved):

bkontur:
@mudigal changed:

image

PR #212, opened 2026-06-17
fix(provider-node): serve /negotiate for freshly registered providers

bkontur:
Freshly registered providers (including after deregister + re-register) have no `ProviderReplayState` yet, so `setup_nonce_counter` returned `None` and every `/negotiate` request 503'd — even with signing fully configured. `/info` gave no hint, since it showed the (healthy) on-chain registration but not the fields that gate signing.

- Bootstrap the nonce counter fresh (start at 1) when there's no on-chain replay state, instead of discarding it.
- Split `NonceCounterUnavailable` out of `SigningUnavailable` so the 503 no longer misreports a missing `--keyfile`.
- Extend `/info` with a `readiness` object (`signing_configured`, `nonce_counter_ready`, `provider_info_loaded`) to diagnose this class of failure without reading logs.


mudigal:
Pushed a follow-up commit (`31968ba`) that builds on this PR to make the node resilient to **post-startup registration**, plus a heads-up on the framing of the original fix.

### Why the extra commit

The PR description says freshly-registered providers have no `ProviderReplayState`, so `setup_nonce_counter` returned `None` and `/negotiate` 503'd. I dug into this and I don't think that's quite the mechanism:

- Registration goes through `register_provider_internal` (used by both the extrinsic at `pallet/src/lib.rs:920` and genesis at `:317`), which inserts **both** maps atomically — `Providers` *and* `ProviderReplayStates::<T>::insert(who, ReplayWindow::default())` (`:3798-3799`).
- An explicit `insert` writes a physical storage key even though the map is `ValueQuery`, so the raw `fetch` in `fetch_replay_hsn` returns `Ok(Some(0))` for any registered provider — not `Ok(None)`.

So the `Ok(None)` branch is only hit for an **unregistered** provider, and that provider is independently blocked by `provider_info == None` → `provider_info_unavailable`. Net: the real-world 503 (which we hit via #195 splitting `start-provider` from `register-provider`) is the provider simply **not being registered when the node starts** — and because both `setup_*` reads are one-shot at startup, registering afterwards never gets picked up without a restart.

### What `31968ba` does

- Replaces the startup-only `setup_provider_info`/`setup_nonce_counter` with a **background reconciler** that polls the chain (first poll immediate, then every 30s). It populates `provider_info` + bootstraps the nonce counter when registration appears, and clears `provider_info` on deregistration — no restart. Settings changes are picked up too.
- `nonce_counter` becomes always-present (`Arc<NonceCounter>`) and now tracks `is_bootstrapped()`; `provider_info` becomes `Arc<RwLock<Option<…>>>` so the reconciler can publish/clear it.
- Kept your `/info` readiness object and `nonce_counter_unavailable` error — `nonce_counter_ready` is now wired to `is_bootstrapped()`, and `/negotiate` keeps a **defensive** `nonce_counter_unavailable` check so it never signs a nonce not derived from on-chain replay state (your error.rs tests pass unchanged; added an integration test for that path).

All 72 unit + integration tests pass; fmt / zepter / clippy `-D warnings` clean.

@bkontur happy to reshape any of this — e.g. drop the defensive check if you'd rather gate purely on registration, or make the poll interval configurable. Wanted to keep your diagnostics and just add the resilience piece.

bkontur:
@mudigal yes, I had similar fix ready `ensure_bootstrapped()` with some interval reading the chain, but it was too complicated to think about it this hours, so I reverted, anyway @danielbui12 is doing the same: https://github.com/paritytech/web3-storage/pull/207 - so we can clean as follow up also

mudigal:
Follow-up commit (`a68d91e`, now merged): **e2e coverage for the reconciler** + a configurable poll interval.

- New `--reconcile-interval-secs` flag (default 30, env `RECONCILE_INTERVAL_SECS`) so the poll cadence can be shortened for tests.
- New papi workflow **`e2e/11-reconciler-resilience.js`**:
  - `11.1` asserts `/info` exposes the readiness object and reports the running provider as signing-configured + registered + nonce-counter-ready.
  - `11.2` bumps the provider's on-chain listed price and asserts a quote reflects the new price after a reconcile tick (settings picked up with no restart), then restores it.
- CI e2e job now starts the provider with `--reconcile-interval-secs 1` so `11.2` converges in ~1s instead of 30s.

Note on scope: the reconciler only tracks the node's *own* account, and the running e2e provider is Alice, so `11` operates on Alice non-destructively (bump → observe → restore). The unregistered → registered and deregistered transitions remain covered by the Rust integration tests (`negotiate_integration.rs`), since the e2e harness starts the node already-registered. All 70 unit + integration tests pass; clippy `-D warnings` clean.

mudigal:
> @mudigal yes, I had similar fix ready `ensure_bootstrapped()` with some interval reading the chain, but it was too complicated to think about it this hours, so I reverted, anyway @danielbui12 is doing the same: #207 - so we can clean as follow up also

@bkontur sounds good — agreed on landing this as the interim fix and converging onto @danielbui12's ChainStateCoordinator in #207 as a follow-up. They're clearly solving the same sub-problem (keep the node's `provider_info` synced from chain behind a `RwLock<Option<ProviderInfo>>`), just two ways:
 - #212 polls on an interval (spawn_chain_reconciler).
 - #207 is event-driven off finalized blocks and also tracks current_block for valid_until stamping.

#207's event-driven approach is the better long-term home - lower latency and none of the per-tick reconnect churn the poller has - so I'd expect its coordinator to replace this reconciler when we merge. Three pieces from here are worth folding in rather than dropping, since #207 focuses on provider_info/valid_until:

 1. the nonce-counter `is_bootstrapped()` gate (so `/negotiate` never signs a nonce not derived from on-chain replay state),
 2. the `/info` readiness object, and
 3. the split error codes (`provider_info_unavailable` / `nonce_counter_unavailable`) instead of the old misleading `signing_unavailable`.

One small thing just added here: `--reconcile-interval-secs` is now constrained to `>= 1` (a 0 busy-looped the chain). That flag goes away once the poller is replaced by the coordinator, so no need to carry it into #207.

 Happy to take the follow-up to graft (1)–(3) onto #207's coordinator once one of these lands.

mudigal (approved):

Issue #149, opened 2026-06-09
Comprehensive codebase audit: security, performance, code quality, and CI gaps


mudigal:
# Comprehensive Codebase Audit

Full audit of the `dev` branch covering security, performance, code quality, TODOs, and CI/config gaps. Findings are grouped by severity across 5 categories.

---

### Security

**S1. Missing signature verification in `delete_data` endpoint**
- [x] `provider-node/src/api.rs:456-478`
- [x] The `admin_signature` field is accepted but explicitly ignored (`let _ = request.admin_signature`). Any user can delete arbitrary bucket data without authentication.
- [x] **Fix**: Verify signature using the auth module before allowing deletion.

**S2. Permissive CORS configuration**
- [x] `provider-node/src/api.rs:80`
- [x] `CorsLayer::permissive()` allows requests from any origin. Combined with S1, this creates a remotely-exploitable attack vector.
- [x] **Fix**: Restrict to trusted origins or make configurable.

### Performance

**P1. Unbounded iteration in `on_finalize` hook**
- [ ] `pallet/src/lib.rs:62-74`
- [ ] Iterates over all expired challenges for a block without bounds. An attacker can create many challenges expiring at the same block, stalling the chain. The `index as u16` cast also silently truncates at 65536.
- [ ] **Fix**: Cap processing per block (e.g., 100), carry remainder to next block.

### CI / Config

**C1. Missing `try-runtime` feature in `pallet-drive-registry`**
- [ ] `storage-interfaces/file-system/pallet-registry/Cargo.toml`
- [ ] Pallet doesn't declare the `try-runtime` feature block. Will break `zepter run check` feature propagation lint.
- [ ] **Fix**: Add `try-runtime` feature with proper propagation.

**C2. Missing `try-runtime` feature in `pallet-s3-registry`**
- [ ] `storage-interfaces/s3/pallet-s3-registry/Cargo.toml`
- [ ] Same issue as C1.

### TODOs blocking functionality

**T1. Checkpoint duty query returns empty vec**
- [ ] `provider-node/src/checkpoint_coordinator.rs:329`
- [ ] `get_active_checkpoint_duties()` has a TODO and returns `Ok(vec![])`. Provider-initiated checkpoints are non-functional.

**T2. Challenge detection returns empty vec**
- [ ] `provider-node/src/challenge_responder.rs:304`
- [ ] Challenge storage query is unimplemented, relies on event detection which is also incomplete.

[...]

PR #215, opened 2026-06-18
Security hardening

### LINKED ISSUE #226: Security hardening for Web3 Storage
_Tracking issue from the cross-repo roadmap plan (2026-06-23)._

Track security hardening (PR #215). Confirm severity — may be P0 if it closes an active vuln.

**Covers PR(s):** #215
**Theme:** UI/Docs/Security/Product · **Category:** Feature
**Priority:** P1 · **Effort:** L (~8 wk) · **Target:** Web3 Mainnet (Dec '26)

> [!NOTE]
> No tracking issue exists; needed pre-mainnet.

RafalMirowski1:
Resolves S1, S2, and S3 flagged in this audit: https://github.com/paritytech/web3-storage/issues/149

S1 — Authenticate POST /delete
  
  The delete endpoint pruned bucket data but never authenticated the caller (let _ = request.admin_signature), so anyone could permanently prune a bucket.
  Now requires an Admin-signed request via the existing check_role header-auth path used by every other mutating endpoint.

  S2 — Configurable CORS
  
  Replaces the hardcoded CorsLayer::permissive() with a --cors-allowed-origins / CORS_ALLOWED_ORIGINS option threaded through ProviderState. Default stays
  permissive (local dev / demos / UIs unchanged); operators can restrict to trusted origins in production.

  S3 — Reject .. in file-system paths

  Adds a shared validate_fs_path helper (absolute, non-empty, no ..) applied to all five fs handlers — previously get/delete/list did no validation.
  Defense-in-depth: fs paths are content-index keys (RocksDB / in-memory), not real filesystem paths, so this isn't an active filesystem-escape.

---

_Closes #226 — cross-repo roadmap tracking._

### DIFF
--- a/docs/design/scalable-web3-storage-implementation.md
+++ b/docs/design/scalable-web3-storage-implementation.md
@@ -1720,8 +1720,10 @@ Rules:
   `Writer` for uploads/commits, `Admin` for delete and other destructive ops.
 - The membership cache uses stale-while-revalidate: if the chain is briefly
   unreachable, cached membership keeps working.
-- When the provider node is launched without auth (`--auth-disabled`), all
-  endpoints are permissive — useful for local demos, never in production.
+- Authentication is enforced by default. The only way to turn it off is the
+  deliberately verbose `--disable-auth-i-know-what-i-am-doing` flag, which makes
+  every endpoint publicly readable and writable. It exists for throwaway local
+  experiments only and must never be used for a real provider.
 
 ### Content-Addressed Storage
 
@@ -1918,7 +1920,7 @@ Delete Data (admin only)
 POST /delete
 Authorization: Web3Storage <pubkey_hex>:<signature_hex>:<timestamp>
   // admin-signed header (same scheme as other mutating endpoints);
-  // when auth is enabled the signer must be an Admin member of the bucket
+  // the signer must be an Admin member of the bucket
 
 Request:
 {
--- a/docs/design/scalable-web3-storage-implementation.md
+++ b/docs/design/scalable-web3-storage-implementation.md
@@ -1916,12 +1916,14 @@ Response (404 Not Found):
 Delete Data (admin only)
 ────────────────────────
 POST /delete
+Authorization: Web3Storage <pubkey_hex>:<signature_hex>:<timestamp>
+  // admin-signed header (same scheme as other mutating endpoints);
+  // when auth is enabled the signer must be an Admin member of the bucket
 
 Request:
 {
   "bucket_id": "0x1234...",
-  "new_start_seq": 10,
-  "admin_signature": "0x..."  // signs {bucket_id, new_start_seq}
+  "new_start_seq": 10
 }
 
 Response (200 OK):
--- a/provider-node/src/api.rs
+++ b/provider-node/src/api.rs
@@ -263,8 +263,18 @@ async fn get_node(
 
 async fn upload_node(
     State(state): State<Arc<ProviderState>>,
+    headers: axum::http::HeaderMap,
     Json(request): Json<UploadNodeRequest>,
 ) -> Result<Json<UploadNodeResponse>, Error> {
+    check_role(
+        &state,
+        &headers,
+        "PUT",
+        request.bucket_id,
+        RequiredRole::Writer,
+    )
+    .await?;
+
     // Decode hash
     let hash_bytes = hex_decode(&request.hash).map_err(|_| Error::InvalidHash {
         expected: request.hash.clone(),
@@ -340,8 +350,18 @@ async fn check_exists(
 
 async fn commit(
     State(state): State<Arc<ProviderState>>,
+    headers: axum::http::HeaderMap,
     Json(request): Json<CommitRequest>,
 ) -> Result<Json<CommitResponse>, Error> {
+    check_role(
+        &state,
+        &headers,
+        "POST",
+        request.bucket_id,
+        RequiredRole::Writer,
+    )
+    .await?;
+
     let data_roots: Vec<H256> = request
         .data_roots
         .iter()
@@ -545,8 +565,11 @@ async fn delete_data(
     headers: axum::http::HeaderMap,
     Json(request): Json<DeleteRequest>,
 ) -> Result<Json<DeleteResponse>, Error> {
-    // Pruning bucket data is destructive and admin-only: require an Admin-signed
-    // request, same as every other mutating endpoint.
+    // Admin-only, unlike the Writer-level deletes in the S3/FS layers: this L0
+    // prune rewrites the underlying MMR (dropping every leaf below
+    // `new_start_seq`), whereas the L1 deletes only drop an index entry and
+    // leave the tree intact. Rewriting the commitment is strictly more
+    // destructive, so it warrants the highest role.
     check_role(
         &state,
         &headers,
--- a/provider-node/src/api.rs
+++ b/provider-node/src/api.rs
@@ -519,10 +519,19 @@ async fn list_buckets(State(state): State<Arc<ProviderState>>) -> Json<ListBucke
 
 async fn delete_data(
     State(state): State<Arc<ProviderState>>,
+    headers: axum::http::HeaderMap,
     Json(request): Json<DeleteRequest>,
 ) -> Result<Json<DeleteResponse>, Error> {
-    // Note: In production, would verify admin_signature
-    let _ = request.admin_signature;
+    // Pruning bucket data is destructive and admin-only: require an Admin-signed
+    // request, same as every other mutating endpoint.
+    check_role(
+        &state,
+        &headers,
+        "POST",
+        request.bucket_id,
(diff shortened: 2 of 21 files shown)


provider-node/src/api.rs
```diff
@@ -48,6 +48,19 @@ pub fn create_router(state: Arc<ProviderState>) -> Router {
...
 
+    // Restrict CORS to the configured origins, or stay permissive when unset.
+    let cors = match &state.cors_allowed_origins {
+        Some(origins) if !origins.is_empty() => {
+            let allowed: Vec<axum::http::HeaderValue> =
+                origins.iter().filter_map(|o| o.parse().ok()).collect();
+            CorsLayer::new()
+                .allow_origin(allowed)
+                .allow_methods(tower_http::cors::Any)
+                .allow_headers(tower_http::cors::Any)
```
franciscoaguirre: Don't we want to restrict methods or headers?

provider-node/src/fs_api.rs
```diff
@@ -46,6 +46,18 @@ fn default_root() -> String {
     "/".to_string()
 }
 
+/// Validate a user-supplied file-system path: absolute, non-empty, and free of
+/// `..` traversal segments.
+fn validate_fs_path(path: &str) -> Result<(), Error> {
```
ilchu, line 51: Can we just convert to `std::fs::Path` and [canonicalize](https://doc.rust-lang.org/std/path/struct.Path.html#method.canonicalize) here?

provider-node/src/api.rs
```diff
@@ -519,10 +532,19 @@ async fn list_buckets(State(state): State<Arc<ProviderState>>) -> Json<ListBucke
 
 async fn delete_data(
     State(state): State<Arc<ProviderState>>,
+    headers: axum::http::HeaderMap,
     Json(request): Json<DeleteRequest>,
 ) -> Result<Json<DeleteResponse>, Error> {
-    // Note: In production, would verify admin_signature
-    let _ = request.admin_signature;
+    // Pruning bucket data is destructive and admin-only: require an Admin-signed
+    // request, same as every other mutating endpoint.
```
ilchu: At the same time in FS and S3 we [require](https://github.com/paritytech/web3-storage/blob/d9b729d49df2a6cdbba4a6bf924ede57f5a11253/provider-node/src/fs_api.rs#L182) deleting to be at least writer role. Is there a reason for tightening here?

provider-node/src/types.rs
```diff
@@ -205,12 +205,13 @@ pub struct MerkleProofData {
 // ─────────────────────────────────────────────────────────────────────────────
 
 /// Request to delete data (admin only).
+///
+/// Authorization is carried in the `Authorization` header (`Web3Storage …`,
+/// verified against the bucket's Admin members), not in the body.
```
ilchu, line 210: On a side note, the auth header format is not documented anywhere besides being constructed ad hoc in unit and integration tests. Maybe I'm missing something, but the overall
```rust
let message = format!("web3storage:{method}:{bucket_id}:{timestamp}");
```
doesn't sound too intuitive to me to go without saying.

ilchu (commented):
Just some small suggestions and thoughts around.

provider-node/src/api.rs
RafalMirowski1: I guess here, already admin signature was required, just not verified. we can change to writer for consistency with FS and S3, wdyt?

provider-node/src/fs_api.rs
RafalMirowski1: paths here are not real paths, but keys, (This also means actual traversal is unlikely, but we can guard against it anyway.) so im not sure about `canonicalize`, I think it needs the path to exist for example.

provider-node/src/types.rs
RafalMirowski1: Addded some documentation

provider-node/src/api.rs
RafalMirowski1: Rsetricted to what we have in our api

provider-node/src/fs_api.rs
ilchu: You're right, I guess it'll have to remain manual.

provider-node/src/api.rs
ilchu: Apparently only this L0 version also prunes the underlying MMR, whereas the L1 S3/FS wrappers only do that with the index, not touching the tree. So probably admin is justified here, but we'd need to state it more clearly in the comment, if at all.

And on another note with the security and authorizing access around endpoints: everything else in the L0 API is completely free for access. So we have all the proper auth roles in S3/FS, but `/node PUT` and `/commit` POST are unauthorized (which should be `RequiredRole::Writer`), and I believe it would make sense to include hardening these into this PR as well.

Then there's also a question of what should be done to `/node` GET, `/read`, `/fetch_nodes` and `/exists` (which are `RequiredRole::Reader` on L1). I guess technically the reader role shouldn't exist at all, as all data is supposed to be client-side encrypted (though that's not completely wired yet), and some provider functionality is meant to be permissionless - such as replica sync. But right now we have a L0/L1 mismatch. So this part is probably better as a followup - I'd appreciate if you could make an issue.

franciscoaguirre (approved):
Looks good after Ilia's comment

provider-node/src/command.rs
```diff
@@ -50,6 +50,7 @@ pub async fn run() -> Result<(), Box<dyn std::error::Error>> {
     let state = match &seed {
         Some(seed) => {
             let mut state = ProviderState::with_seed(storage, seed)?;
```
bkontur: @RafalMirowski1 I would clean up this initialization: `ProviderState::with_seed` vs `ProviderState::new` -> I would add dedicated `ProviderState::with_provider_id` (also deduplicate with_seed and new internally as a follow up), and on the top of that extend both with `cors_allowed_origins`. Also we are doing here lots of setters `state.* = *` for auth, that could be also refactored to `state.setAuthConfig` or something - another follow-up

provider-node/src/api.rs
RafalMirowski1: Issue created here: https://github.com/paritytech/web3-storage/issues/228. Both `/node` PUT and `/commit` POST now require writer role and added some comment why admin makes sense.

provider-node/src/command.rs
RafalMirowski1: Added `ProviderState::with_provider_id` added `with_cors_origins`, `set_auth_config` methods and `configure_state` helper.

docs/design/scalable-web3-storage-implementation.md
```diff
@@ -1916,12 +1916,14 @@ Response (404 Not Found):
 Delete Data (admin only)
 ────────────────────────
 POST /delete
+Authorization: Web3Storage <pubkey_hex>:<signature_hex>:<timestamp>
+  // admin-signed header (same scheme as other mutating endpoints);
+  // when auth is enabled the signer must be an Admin member of the bucket
```
eskimor: When is auth ever not enabled?

RafalMirowski1: there's a flag `--enable-auth` on provider node, I guess we can run without enabling for local runs, testing, demos etc

eskimor: auth should definitely be the default and it should not run without it. The system should be secure by default. We can have a flag --disable-auth-i-know-what-i-am-doing , but we should really not do the opposite. Deployments should be by default secure or non-functional. (Strong preference on the former of course.)

RafalMirowski1: that makes sense :) introduced `--disable-auth-i-know-what-i-am-doing` flag, and auth enabled is the default now, are you fine with me using this flag here for ci, and adjust tests in a followup?

eskimor (approved):
Thanks!

docs/design/scalable-web3-storage-implementation.md
eskimor: Looks good. Thanks! Let's keep and enforce this policy. Default should be secure or non-functional. No insecure defaults ever.

bkontur: haha, I asked @danielbui12 to remove that `--enable-auth` arg some time ago here: https://github.com/paritytech/web3-storage/issues/88#issuecomment-4623356915


@RafalMirowski1 why do we want to run `--disable-auth-i-know-what-i-am-doing` for CI? Especially in CI, we should run everything, every possible check, no by-passing anything.

I would remove `disable_auth_i_know_what_i_am_doing` - who will be really using this?

RafalMirowski1: i guess you're right we don't wanna use this for CI, but i think some tests/demos don't send auth headers now so they would fail, and i wanted to change that as a followup, but I can do this here as well?

bkontur: ok, ok, let's do follow up with removed arg and correct headers stuff

PR #217, opened 2026-06-18
Bound the Challenges storage map

RafalMirowski1:
Convert the value of `Challenges[n]`to `BoundedVec<Challenge<T>, T::MaxChallengesPerBlock>`.
Flagged by audit here: https://github.com/paritytech/web3-storage/issues/149 as P1.

### DIFF
--- a/runtimes/web3-storage-paseo/src/storage.rs
+++ b/runtimes/web3-storage-paseo/src/storage.rs
@@ -89,6 +89,7 @@ impl pallet_storage_provider::Config for Runtime {
     type CheckpointReward = CheckpointReward;
     type CheckpointMissPenalty = CheckpointMissPenalty;
     type MaxBucketsPerMember = ConstU32<1000>;
+    type MaxChallengesPerBlock = ConstU32<1024>;
     type DeregisterAnnouncementPeriod = DeregisterAnnouncementPeriod;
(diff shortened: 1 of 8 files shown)


runtimes/web3-storage-paseo/src/storage.rs
```diff
@@ -89,6 +89,7 @@ impl pallet_storage_provider::Config for Runtime {
     type CheckpointReward = CheckpointReward;
     type CheckpointMissPenalty = CheckpointMissPenalty;
     type MaxBucketsPerMember = ConstU32<1000>;
+    type MaxChallengesPerBlock = ConstU32<1024>;
```
franciscoaguirre, line 92: Where did we get this number?

RafalMirowski1: it's random (or a placeholder sounds nicer :)), I wasn't exactly sure how to arrive at a reasonable bound for this

ilchu (commented):
I've done a very similar thing in my challenges [PR](https://github.com/paritytech/web3-storage/pull/125/changes/f255275ac16e2688ff058b8ffa54e6948ca3b5b9), but also added weight bounds as a hook to keep things bounded. I guess it' my fault for that PR being huge, but otherwise we can close this one.

Issue #175, opened 2026-06-15
[Tooling] Add a `storage-cli` tooling crate (on-chain + off-chain ops), starting with `stress-test upload`


bkontur:
## Motivation

We've accumulated a bunch of useful commands and snippets for exercising on-chain
and off-chain storage scenarios (registering providers, opening agreements,
uploading data, etc.), currently scattered across examples and ad-hoc scripts.
Consolidate them into one ergonomic binary so contributors and CI can drive both
layers from a single tool.

## Proposal

Add a new workspace crate **`storage-cli`** under **`utils/storage-cli`**, built on
the existing `storage-client` SDK.

- New crate `utils/storage-cli`, added to `[workspace].members` in the root `Cargo.toml`.
  (`utils/` follows the Polkadot SDK convention of keeping developer tooling under
  `substrate/utils/`.)
- `clap`-based CLI (derive API), one subcommand per operation, with shared global flags:
  - `--chain-rpc <URL>` (default `ws://127.0.0.1:2222`)
  - `--provider-url <URL>` (provider HTTP endpoint; default `http://127.0.0.1:3333`)
  - `--suri <SURI>` / `--keyfile <FILE>` for the signing account
- Reuse `storage-client` (`StorageUser`, discovery, substrate helpers) instead of
  re-implementing chain/HTTP access.

## First command: `stress-test upload`

Simplest case — "upload data to one provider".

**Behavior**

1. Resolve target buckets from chain for the (client, provider) pair: read the
   client's buckets (`MemberBuckets[client]`) and keep those that have an active
   `StorageAgreements[bucket][provider]` entry (i.e. an agreement between this
   client and the given provider).
2. If no matching buckets exist -> exit with a clear error. Do **not** create any
   bucket or agreement.
3. Upload generated data to each selected bucket via the provider's HTTP endpoint
   (`StorageUser::upload`).

**Params**

- `--provider <AccountId>` (required) — the provider to target.
- `--max-buckets-to-write <N>` (optional) — cap the number of buckets written to;
  default = all matching buckets.
- `--size <BYTES>` (optional, suggested) — amount of data to generate per upload,
  with a sensible default.

## Acceptance criteria

- [ ] `storage-cli` crate compiles as a workspace member; `cargo run -p storage-cli -- --help`
      lists the `stress-test` subcommand.
- [ ] `stress-test upload --provider <id>` uploads to all matching buckets;
      `--max-buckets-to-write N` caps the count.
[...]

Issue #214, opened 2026-06-18
[Deployment] Design telemetry and metrics for on-chain and off-chain components


bkontur:
**Hints:**
- for on-chain, it will be most probably deployed directly to the AH, so no specially care
- for off-chain, add possibility for provider-node to expose metrics to grafana/prometheus 
  - setup and use the same mechanism (crates from PolkadotSDK) for metrics as we have for PolkadotSDK nodes/binaries
- investigate and add relevant metrics to the provider node

PR #220, opened 2026-06-19
Add storage-cli tooling crate with `stress-test upload`

darwinsubramaniam:
## Summary

Adds `utils/storage-cli`, a clap-based operator CLI built on the `storage-client` SDK, starting with a `stress-test upload` subcommand.

## Changes

- New workspace member `utils/storage-cli`; promotes `clap` to a workspace dependency (version only, features set per-crate).
- `stress-test upload` resolves target buckets from chain (`MemberBuckets[account]` intersected with buckets that have a `StorageAgreements[bucket][provider]` entry) and uploads generated data over the provider HTTP API. It never creates buckets or agreements, and errors clearly when no matching buckets exist.
- Reuses `AdminClient` (chain reads) and `StorageUserClient` (HTTP upload); no duplicated chain or HTTP logic.
- Adds `utils/storage-cli/README.md` documenting usage.

Related to #175 

## Notes for reviewers

- Opened as a **draft** for review.
- See `utils/storage-cli/README.md` for invocation examples.


utils/storage-cli/src/shared.rs
```diff
@@ -0,0 +1,94 @@
+use crate::cli_args::GlobalArgs;
```
bkontur: I would change shared.rs to `common/mod.rs` - there will come more stuff for sure

utils/storage-cli/src/scenarios/mod.rs
```diff
@@ -0,0 +1 @@
+pub mod stress_test;
```
bkontur: maybe rename `scenarios` -> `commands`

crates/utils/storage-cli/src/cli.rs
```diff
@@ -0,0 +1,78 @@
+//! Command-line argument parsing for the storage CLI.
```
bkontur, line 3: I would rename cli_args.rs to cli.rs

utils/storage-cli/src/cli_args.rs
```diff
@@ -0,0 +1,78 @@
...
+
+#[derive(Debug, Subcommand)]
+pub enum Command {
+    /// Stress-test operations against a provider.
+    #[command(subcommand)]
+    StressTest(StressTest),
+}
+
+#[derive(Debug, Subcommand)]
+pub enum StressTest {
```
bkontur: StressTest + UploadArgs could better go to the `commands/mod.rs` or `commands/stress_ test.rs` directly (let's do former for now)

utils/storage-cli/src/cli_args.rs
```diff
@@ -0,0 +1,78 @@
...
+    /// Upload generated data to every bucket the account already has an
+    /// agreement with the given provider for.
+    Upload(UploadArgs),
```
bkontur: Maybe just `Upload` -> `ProviderUpload`?
```suggestion
    ProviderUpload(UploadArgs),
```

utils/storage-cli/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+[dependencies]
+storage-client = { workspace = true }
+clap = { workspace = true, features = ["derive", "env"] }
+sp-core = { workspace = true, features = ["std"] }
+sp-runtime = { workspace = true, features = ["std"] }
+subxt-signer = { workspace = true }
+tokio = { workspace = true }
+tracing-subscriber = { workspace = true, features = ["env-filter"] }
+anyhow = "1"
+hex = "0.4"
```
bkontur: move to workspace deps
```suggestion
anyhow  = { workspace = true }
hex  = { workspace = true }
```

crates/utils/storage-cli/src/main.rs
```diff
@@ -0,0 +1,25 @@
+//! `storage-cli` — operator CLI for scalable Web3 storage.
```
bkontur, line 3: @darwinsubramaniam please, tell you assistant to align all file headers (licences) with the rest of the codebase

bkontur:
@darwinsubramaniam also it would be cool to add this tool and all its commands to the CI `integration-tests.yml` as e.g. `utils-integration-tests`

utils/storage-cli/src/cli_args.rs
darwinsubramaniam: `Provider` name here maybe conflict with the role of `provider-node` .. would it be more distinguishable if call it as `ClientUploud`. Client here refer to the account which wants to upload data into the provider.

darwinsubramaniam:
@bkontur , Ok will add that part as well.

utils/storage-cli/src/cli_args.rs
darwinsubramaniam: @bkontur  - please review the comment on the naming

bash command : storage-cli  -h 
image

bash command : storage-cli stress-test
image

the do you want the cli subcommand under the stress-test to be `upload` or `provider-upload`? For now i kept it for the end user as `upload` as it was stated in the Issue #175 . If the `provider-upload` makes sense here I am ok to change the override.

utils/storage-cli/src/commands/stress_test.rs
```diff
@@ -0,0 +1,136 @@
...
+        let data_root = user
+            .upload(*bucket, &payload, ChunkingStrategy::default())
+            .await
+            .with_context(|| format!("upload to bucket {bucket} failed"))?;
+        println!(
+            "  bucket {bucket}: uploaded {} bytes, data_root 0x{}",
+            payload.len(),
+            hex::encode(data_root.as_bytes()),
+        );
+    }
```
bkontur: This is the most important part of this stress-tess, based on the configuration, we should have possibilities:
- configure number of users (1..N) (default 1)
- configure max payload size (by default 0.5 MiB) - (it will generate random payloads)
- configure mode(s): ParallelUsers(1..N) + ParallelUploads(1..X)

The idea is to have ability specify different scenarios just by configuration:
- 1 user with 1000 sequence uploads with max size 0.5 MiB
- 1 user with 100 parallel uploads with max size 0.5 MiB
- 10 parallel users with 100 parallel uploads with max 0.5 MiB
- (just a note: I know, we can run also run the stress-test binaries in parallel)



```suggestion 
    // TODO: list of parallel users
    // Off-chain HTTP uploads (no chain, no signer). Constant-fill payload,.
    let user = StorageUserClient::new(config).context("failed to construct provider client")?;
    // TODO: dynamic random content with max size by configuration
    let payload = vec![0x42; args.size];

    // TODO": make this run in parallel in case of configuration
    for bucket in &selected_buckets_id {
        let data_root = user
            .upload(*bucket, &payload, ChunkingStrategy::default())
            .await
            .with_context(|| format!("upload to bucket {bucket} failed"))?;
        println!(
            "  bucket {bucket}: uploaded {} bytes, data_root 0x{}",
            payload.len(),
            hex::encode(data_root.as_bytes()),
        );
    }
```

utils/storage-cli/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+sp-core = { workspace = true, features = ["std"] }
+sp-runtime = { workspace = true, features = ["std"] }
+subxt-signer = { workspace = true }
+tokio = { workspace = true }
+tracing-subscriber = { workspace = true, features = ["env-filter"] }
+anyhow = { workspace = true }
+hex = { workspace = true }
+
+[dev-dependencies]
+tempfile = "3"
```
bkontur: please as workspace deps

utils/storage-cli/src/commands/stress_test.rs
```diff
@@ -0,0 +1,136 @@
...
+    pub size: usize,
+}
+
+/// Upload generated data to every bucket the account already has an agreement
+/// with `--provider` for.
+///
+/// This resolves targets from chain (`MemberBuckets[account]` ∩ buckets with a
+/// `StorageAgreements[bucket][provider]` entry) and never creates buckets or
+/// agreements — if nothing matches, it errors out.
+pub async fn upload(global: &GlobalArgs, args: &UploadArgs) -> Result<()> {
```
bkontur: @darwinsubramaniam at the end, it should at least print some metrics (at least some times, size uploaded, ...). On the other hand (outside of the scope of this PR), we will collect internal statistics https://github.com/paritytech/web3-storage/issues/214

danielbui12: Ideally, it should support e2e flow: ensure provider registered ->  create bucket -> upload data. You can see many examples at `./examples/papi/e2e`

.github/workflows/integration-tests.yml
```diff
@@ -547,6 +547,131 @@ jobs:
...
+
+      - name: Wait for provider health
+        uses: ./.github/actions/wait-for-provider-health
+
+      # Seed: //Bob negotiates terms and opens a bucket + primary agreement
+      # against the //Alice provider, giving the CLI something to upload to.
+      - name: Seed bucket + agreement (//Bob → //Alice provider)
+        run: |
+          cargo run --release -p storage-client --example complete_workflow \
+            ws://127.0.0.1:2222 http://127.0.0.1:3333 //Bob
```
danielbui12, line 699: It is better to implement the logic directly in the crate/module. These examples may change over time, and duplicating logic between the examples and the crate/module can increase CI execution time and maintenance overhead

utils/storage-cli/src/commands/stress_test.rs
danielbui12: For cases involving many users where a many extrinsics need to be invoked, you can consider using pallet_utility to submit batched transactions. This improves overall execution efficiency. However, keep in mind the block weight and size limits when constructing large batches.

.github/workflows/integration-tests.yml
```diff
@@ -547,6 +547,131 @@ jobs:
...
+      - name: Wait for parachain blocks
+        uses: ./.github/actions/wait-for-parachain
+        with:
+          log-file: /tmp/omni-node.log
+
+      - name: Register provider on-chain
+        run: |
+          echo "//Alice" > /tmp/alice-key && chmod 600 /tmp/alice-key
+          cargo run --release -p storage-client --example register_provider \
+            ws://127.0.0.1:2222 http://127.0.0.1:3333 /ip4/127.0.0.1/tcp/3333 /tmp/alice-key
```
danielbui12, line 683: Ref: https://github.com/paritytech/web3-storage/pull/220/changes#r3471870419
Recently I dropped all provider registration jobs because new module  chain_state_coordinator will keep the provider node up-to-date with runtime https://github.com/paritytech/web3-storage/pull/207

utils/storage-cli/src/commands/stress_test.rs
bkontur: > Ideally, it should support e2e flow: ensure provider registered -> create bucket -> upload data. You can see many examples at `./examples/papi/e2e`

hard to say, e2e could be another command, I am sure that for other issues we will want to focus on optimizing just particular stuff - like fast upload, fast download, so we would want to stress just those parts, not the whole flow

utils/storage-cli/src/commands/stress_test.rs
bkontur: > For cases involving many users where a many extrinsics need to be invoked, you can consider using pallet_utility to submit batched transactions. This improves overall execution efficiency. However, keep in mind the block weight and size limits when constructing large batches.

@danielbui12 what do you mean by using pallet_utility here? This is just stretching provider upload RPC, no on-chain call, no transaction.

utils/storage-cli/src/commands/stress_test.rs
darwinsubramaniam: Hi @bkontur  and @danielbui12 
I hope i get clear idea here.

The upload function here will check if there is any agreement belongs to the requester on the targeted provider. If found upload , if non just bail from the test as no agreement made.  So the agreement is an assumption , as pre-setup for now. 

Do you want this pre-setup stage to be also automated?  

Sorry for the late reply, I am current in my master's  exam period. So I might be slow  to response and get it moving, but I am keen on completing this PR

.github/workflows/integration-tests.yml
darwinsubramaniam: #https://github.com/paritytech/web3-storage/pull/220#discussion_r3491337317
@bkontur  and @danielbui12 

This is the pre-setup stage , which i just mentioned . So this particular pre-setup move it into stage the Upload function as default ?

darwinsubramaniam:
@bkontur @danielbui12 quick idea I wanted to float

Now that `@web3-storage/sdk` exists, what do you think about moving `storage-cli` (and its `stress-test` subcommand) off Rust and onto the TS SDK instead? 

Rationale: right now we maintain two parallel HTTP clients (Rust `storage-client` and the TS SDK) that both talk to the same provider API, and they can silently drift out of sync 

e.g. the Rust client still doesn't sign requests per `provider-node`'s auth scheme, while the TS SDK already does. 

Consolidating on one client would mean one source of truth for anything upstream (auth, chunking, agreement/upload flow), instead of having to remember to update both.

Not proposing we do this right now ,mostly want to know if the direction makes sense to you two before I look into what it'd actually take (there'd be some gaps to fill on the TS side, nothing that looks like a blocker so far). Thoughts?

danielbui12:
@darwinsubramaniam we will have both, we prefer the TS sdk to rust sdk. but I have been working on rust sdk now https://github.com/paritytech/web3-storage/pull/239

IMO, a stress test tool must generate more load than the target can handle, and measure accurately. Under saturation NodeJS event loop delays response timestamps, inflating recorded latency and garbage collector is a minor factor. Rust is still solid choice, no GC jitter, multi-core without IPC overhead, predictable tail behavior at the generator side.

darwinsubramaniam:
@danielbui12 
I made the item work now using no auth provider for now . Please review if my overall understanding is correct scope . 

For one of your review making stress test able to register and create agreement can I just add a flag --create-agreement --start-provider to make it optional argument

crates/utils/storage-cli/src/commands/stress_test.rs
```diff
@@ -0,0 +1,343 @@
...
+        &account_ss58,
+        &provider_hex,
+        &args.provider,
+        &global.chain_rpc,
+    )
+    .await?;
+    if let Some(max) = args.max_buckets_to_write {
+        buckets.truncate(max);
+    }
+
```
x3c41a, line 316: `--max-buckets-to-write 0` slips through here (the empty-buckets check runs before this truncate), and `bucket_for` later does `idx % 0` and panics. Make the flag `NonZeroUsize` like the other count flags, or reject 0.

crates/utils/storage-cli/Cargo.toml
```diff
@@ -0,0 +1,25 @@
+[package]
+name = "storage-cli"
+version = "0.1.0"
+authors.workspace = true
+edition.workspace = true
+license.workspace = true
```
x3c41a: workspace license is GPL-3.0-only but the source headers here are Apache-2.0 - match storage-indexers:
```suggestion
license = "Apache-2.0"
```

crates/utils/storage-cli/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+repository.workspace = true
+description = "Operator CLI for scalable Web3 storage (stress-test and ops tooling)"
+
+[dependencies]
+storage-client = { workspace = true }
+clap = { workspace = true, features = ["derive", "env"] }
+sp-core = { workspace = true, features = ["std"] }
+sp-runtime = { workspace = true, features = ["std"] }
+subxt-signer = { workspace = true }
```
x3c41a: unused - signing goes through `storage_client::Signer`:
```suggestion
```

.github/workflows/integration-tests.yml
```diff
@@ -604,6 +604,133 @@ jobs:
...
+  utils-integration-tests:
+    name: Utils Integration Tests
+    runs-on: parity-large
+    timeout-minutes: 30
+    needs: [set-image, build]
+    container:
+      image: ${{ needs.set-image.outputs.CI_IMAGE }}
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6
```
x3c41a: only job in the file on the old pin - align with the rest:
```suggestion
        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
```

.github/workflows/integration-tests.yml
```diff
@@ -604,6 +604,133 @@ jobs:
...
+          persist-credentials: false
+
+      - name: Load common environment variables via env file
+        run: cat .github/env >> $GITHUB_ENV
+
+      - name: Install just
+        run: cargo install just --locked || true
+
+      - name: Cache Polkadot SDK binaries
+        uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5
```
x3c41a: same - the rest of the file uses the newer pin:
```suggestion
        uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v5
```

.github/workflows/integration-tests.yml
```diff
@@ -604,6 +604,133 @@ jobs:
...
+            .bin/chain-spec-builder
+          key: polkadot-sdk-${{ env.POLKADOT_SDK_VERSION }}
+
+      - name: Download Polkadot SDK binaries
+        if: steps.sdk-cache.outputs.cache-hit != 'true'
+        run: just polkadot_version="$POLKADOT_SDK_VERSION" download-polkadot-sdk-binaries
+
+      - name: Resolve chain-spec script
+        run: |
+          RUNTIME=web3-storage-paseo
```
x3c41a: other jobs use the workflow env instead of hardcoding:
```suggestion
          RUNTIME=$INTEGRATION_TEST_RUNTIME
```

crates/utils/storage-cli/src/commands/stress_test.rs
```diff
@@ -0,0 +1,343 @@
...
+    pub max_concurrency: usize,
+}
+
+/// Pick the target bucket for the `global_idx`-th upload of the whole run,
+/// round-robin so load spreads evenly across the selected buckets.
+fn bucket_for(global_idx: usize, buckets: &[BucketId]) -> BucketId {
+    buckets[global_idx % buckets.len()]
+}
+
+/// Generate a payload of exactly `size` random bytes.
```
x3c41a, line 75: nit: `--max-payload-size` is an exact size, not a cap (README documents it as exact) - either rename to `--payload-size` or actually randomize up to the max

darwinsubramaniam: You are right, the arg name `--payload-size` is better. I will avoid using the randomiser up to max; the stress test will not be identical from one run to another.

dr333ws:
Hey @darwinsubramaniam 

What's the status of this PR? It has been in the review for more than 3 month...
Do you need anything from us?

darwinsubramaniam:
@dr333ws , I dont have any permission to merge this PR. The PR being approved but dont know if I call for the merge ?

dr333ws:
I approved the workflow, @darwinsubramaniam could you resolve/close the remaining comments or ping relevant people if you're unsure.

Issue #133, opened 2026-06-04
[Story] dApps integration example (agreement, upload, download)


bkontur:
**Hints**
- we already have web3-storage SC precompile ready  and pallet-revive in the w3s testnet runtimes (no need for asset hubs)
- add to the landing page + new page/ui -> dApps example
- showcase simple dApp UI using smart contract with integration to Web3 Storage
- covering cases:
  - signed user does not have storage agreement
  - signed user has agreement and can store some data
  - show case upload/download

# TODO

- [ ] prepare simple design/description first
- [ ] implement approved design

PR #229, opened 2026-06-23
Photos: part 2

franciscoaguirre:
Part of #133


user-interfaces/photos/scripts/photos-flow.ts
```diff
@@ -111,13 +140,164 @@ async function main() {
...
+    for (const u of uploaded) {
+      const entry = fresh.find((e) => e.path === u.path);
+      assert.ok(entry, `provider dropped ${u.path}`);
+      assert.strictEqual(toHex(entry.dataRoot).toLowerCase(), toHex(u.dataRoot).toLowerCase(), `provider-served content for ${u.path} != the client's uploaded bytes`);
+    }
+    const recomputed = (toHex(metadataMerkleRoot(fresh)) as `0x${string}`).toLowerCase();
+    // Read the anchor back; retry briefly in case the best-block view lags the
+    // just-included setRoot (a still-zero rootCid here means the write missed).
+    let anchored = await readLibraryOf(api, deployed.addressBytes, substrateToH160(user.publicKey), user.address, abi);
+    for (let i = 0; i < 5 && /^0x0+$/.test(anchored.rootCid); i++) {
```
danielbui12: Readability improvement:  as personal habit, I always put deterministic value in a constant variable.
Similar to line 264.

```suggestion
    const ANCHOR_RETRY_BLOCKS = 5;
    
    for (let i = 0; i < ANCHOR_RETRY_BLOCKS && /^0x0+$/.test(anchored.rootCid); i++) {
```

user-interfaces/photos/scripts/photos-flow.ts
```diff
@@ -111,13 +140,164 @@ async function main() {
...
+      await waitForNextBlock(papi);
+      editAnchored = await readLibraryOf(api, deployed.addressBytes, substrateToH160(user.publicKey), user.address, abi);
+    }
+    const idxAfterEdit = await indexRoot(providerUrl, bucketId);
+    console.log(`  recomputed=${recomputedAfterEdit}`);
+    console.log(`  on-chain  =${editAnchored.rootCid.toLowerCase()}`);
+    console.log(`  index_root=${idxAfterEdit.metadataMerkleRoot.toLowerCase()}`);
+    assert.strictEqual(recomputedAfterEdit, editAnchored.rootCid.toLowerCase(), "recomputed root != on-chain rootCid (post-edit)");
+    assert.strictEqual(recomputedAfterEdit, idxAfterEdit.metadataMerkleRoot.toLowerCase(), "recomputed root != provider index_root (post-edit)");
+    assert.notStrictEqual(editAnchored.rootCid.toLowerCase(), rootBeforeEdit, "on-chain rootCid did not move after the edit");
```
danielbui12: Readability improvement:  the "recompute -> read anchor -> retry loop -> assert vs on-chain + index_root" block is duplicated as above block (line from 206 to 219). we could have a local function  to abstract this logic. i.e, `verifyAnchor(..., { expectedEntries, isStale }) returning { recomputed, anchored, idx }` .

packages/core/src/merkle.ts
```diff
@@ -0,0 +1,128 @@
+// SPDX-License-Identifier: GPL-3.0-only
```
bkontur, line 1: Didn't check in deep, but looks like something generic that others could also need/reuse, so maybe good candidate to move to the SDK

bkontur (approved):

PR #234, opened 2026-06-25
Photos: part 3

franciscoaguirre:
Part of #133


user-interfaces/photos/src/lib/chain-client.ts
danielbui12, line 1: we have shared NetworkPicker component in `user-interfaces/shared/network-picker` ; just wonder this is a standalone project or should be wired to shared ui?

user-interfaces/landing/index.html
```diff
@@ -219,6 +219,11 @@ <h2>Drive</h2>
         <h2>S3 Storage</h2>
         <p>Amazon S3-compatible object storage on the decentralized web. Create buckets, manage objects, and browse prefixes.</p>
       </a>
+      <a href="./photos/" class="card" data-app="photos" data-testid="landing-card-photos">
```
bkontur, line 222: @franciscoaguirre 👍 exactly

user-interfaces/landing/index.html
```diff
@@ -219,6 +219,11 @@ <h2>Drive</h2>
...
+        <div class="icon">&#x1F5BC;</div>
+        <h2>Photos</h2>
+        <p>A photo library backed by decentralized storage. Organize albums, upload and view photos, with the album-tree root anchored on-chain.</p>
```
bkontur: I would just add here some note, that it is also example/demonstration of integration of smart contract and storage solution.

```suggestion
        <p>A photo library backed by decentralized storage. Organize albums, upload and view photos, with the album-tree root anchored on-chain.</p>
        <p>TODO: example of integration dApps, smart contract and storage</>
```

user-interfaces/photos/src/contract/photos-abi.ts
```diff
@@ -0,0 +1,88 @@
...
+//
+// ABI for the Photos contract, transcribed from `contracts/Photos.sol` (and the
+// `PrimitiveAgreementTerms` tuple from `contracts/IDriveRegistry.sol`). The UI
+// needs the ABI for viem encode/decode; it deliberately does NOT import the
+// generated `src/contract/Photos.json` because that artifact is produced by
+// `build:contract` (solc + resolc) and is git-ignored, so it isn't present in a
+// plain checkout or in CI. The contract `bin` (needed only for the optional
+// dev-deploy fallback, M8) still comes from `Photos.json` at that point.
+//
+// Keep in sync with `contracts/Photos.sol` if the contract interface changes.
```
bkontur: @franciscoaguirre just really out of curiosity, is this the standard way? Does not PAPI or other tool has support for `.sol` contract? I mean, can we generate this somehow from `.sol` files? Doesn't tool like this exist?

user-interfaces/photos/src/lib/chain-client.ts
bkontur: yes, we want to wire it to the actual w3s ui and landing page, which is good and it uses shared network config.

One thing, lots of these stuff is also in https://github.com/paritytech/web3-storage/blob/dev/user-interfaces/provider/src/lib/chain-client.ts#L33 - maybe worth to refactor to shared stuff as generic?

user-interfaces/photos/src/contract/photos-abi.ts
franciscoaguirre: resolc already emits the ABI, I just wasn't extracting it to a TS file from it. I'll see if I can do that in extract.mjs

user-interfaces/shared/chain-client/src/index.ts
```diff
@@ -0,0 +1,120 @@
...
+// ─────────────────────────────────────────────────────────────────────────────
+
+let client: PolkadotClient | null = null
+let api: ParachainApi | null = null
+let currentEndpoint: string = ''
+
+export const clientReady$ = new BehaviorSubject<boolean>(false)
+
+export async function connectToChain(endpoint: string): Promise<PolkadotClient> {
+  if (client && currentEndpoint === endpoint) return client
```
bkontur: My assistant proposes that we should clean up, in case changing endpoints: with `disconnecFromChain()` or `client.destroy` - not sure which one:
```suggestion
  if (client) {
    if (currentEndpoint === endpoint) return client
    // Endpoint changed: tear down the old client first, or its WebSocket
    disconnectFromChain()
  }
```

user-interfaces/photos/src/utils/cn.ts
```diff
@@ -0,0 +1,8 @@
+// SPDX-License-Identifier: GPL-3.0-only
+
+import { clsx, type ClassValue } from 'clsx'
+import { twMerge } from 'tailwind-merge'
+
+export function cn(...inputs: ClassValue[]) {
```
bkontur, line 11: maybe some oneliner docs, what is this and why, `cn` does not say much :)

user-interfaces/photos/src/state/chain.state.ts
```diff
@@ -0,0 +1,71 @@
...
+  try {
+    await connectToChain(ep)
+
+    // Apply chain-derived config (the SS58 prefix) before serving as connected.
+    const chainProps = await getChainProperties()
+    await configureFromChain(chainProps)
+
+    connectionStatus$.next('connected')
+
+    // finalizedBlock$ replays the current head into blockNumber$ on subscribe,
```
bkontur, line 53:
```suggestion
    // Drop the previous subscription or they stack.
    blockUnsubscribe?.()
    // finalizedBlock$ replays the current head into blockNumber$ on subscribe,
```

PR #235, opened 2026-06-25
Photos: part 4

franciscoaguirre:
Part of #133


user-interfaces/photos/src/lib/fs-client.ts
danielbui12, line 1: Awesome work! this could be refactored to a shared folder or dedicated sdk for re-usable purpose

user-interfaces/photos/src/lib/photos-providers.ts
```diff
@@ -0,0 +1,98 @@
...
+  maxCapacity: bigint
+  minDuration: number
+  maxDuration: number
+}
+
+/**
+ * Walk `StorageProvider.Providers` and return every registered provider with
+ * the fields the picker needs, sorted by free capacity descending.
+ */
+export async function listProviders(api: ParachainApi): Promise<PhotosProvider[]> {
```
danielbui12, line 41: We have runtime api StorageProvider::query_find_matching_providers that can filter based on user's requirement & provider's expectation

user-interfaces/photos/scripts/lib/papi.ts
```diff
@@ -0,0 +1,191 @@
...
+  mnemonicToEntropy,
+  ss58Address,
+} from "@polkadot-labs/hdkd-helpers";
+import { parachain } from "@polkadot-api/descriptors";
+import type { ParachainApi } from "@web3-storage/papi";
+
+const devMiniSecret = entropyToMiniSecret(mnemonicToEntropy(DEV_PHRASE));
+const deriveSr25519 = sr25519CreateDerive(devMiniSecret);
+
+export interface Signer {
```
bkontur: I would check and refactor/reuse from packages/core or packages/layer0, we have there similar stuff e.g.:
```
/**
 * A signer bundle: the PAPI `PolkadotSigner` plus the identity details every
 * caller ends up needing alongside it. Pallet wrappers take this shape so
 * call sites never juggle `(signer, address, publicKey)` triples.
 */
export interface ChainSigner {
```

bkontur (approved):

Issue #64, opened 2026-05-04
[Client] Implement dedicated SDK Checklist


danielbui12:
### High Impact — Client SDK stubs (`client/src/`)

~30 TODOs where chain interaction is placeholder logic. Use `client/src/storage_user.rs` and `client/src/substrate.rs` as reference patterns for real `subxt` calls.

| File | TODOs | Examples |
|---|---|---|
| `client/src/admin.rs` | ~13 | `create_bucket`, `establish_agreement`, `submit_checkpoint` |
| `client/src/provider.rs` | ~10 | `register_provider`, `update_settings`, `get_all_providers` |
| `client/src/discovery.rs` | ~5 | All 5 runtime API calls unimplemented |

---

### Medium — Provider Node gaps

| File | What's missing |
|---|---|
| `provider-node/src/checkpoint_coordinator.rs` | Query chain for buckets where this provider is primary |
| `provider-node/src/challenge_responder.rs` | Proper storage query for `Challenges` |

---

### Medium — S3 interface

| File | What's missing |
|---|---|
| `storage-interfaces/s3/client/src/substrate.rs` | Pagination for `list_objects` |

bkontur:
I would close one of this or https://github.com/paritytech/web3-storage/issues/68 - looks like duplicate.

After experiences with Bulletin SDK, I would be very careful with SDK, for products primary we need good TS-like SDK. 
I would not put too much effort to Rust SDK one for now.

danielbui12:
@bkontur  this one is a parent issue as a checklist, going with many sub-issues

bkontur:
Let's not waste much time on Rust SDK, we will probably need TS SDK for dApps more.

PR #239, opened 2026-06-25
feat: migrate storage-client and provider-node to typed subxt bindings via storage-subxt

### LINKED ISSUE #237: Expose interfaces via a separate `RuntimeApi`
There's some generated code across provider and client that does things like:

```rust
// DoubleMap key layout:
//   16 (pallet hash) + 16 (storage hash)
//   + 16 (blake2_128 of deadline) + 4 (BlockNumber u32 LE)
//   + 8 (twox64 of index)         + 2 (u16 index LE)
//   = 62 bytes total.
let key_bytes = &kv.key_bytes;
if key_bytes.len() < 62 {
    continue;
}
let deadline = u32::from_le_bytes(key_bytes[48..52].try_into().unwrap_or([0; 4]));
let index = u16::from_le_bytes(key_bytes[60..62].try_into().unwrap_or([0; 2]));

```
So it's better if we address it:
> hmm, this direct storage key parsing is kind of weak and fragile, for example what if the pallet changes the format.
> 
> I would do this call over RuntimeApi - expose versioned runtime api, if the format or struct changes, we just bump runtime api and the provider will fail here - please fix or create follow-up issue, do we use this storage item in provider anywhere else?
> 
> _Originally posted by @bkontur in https://github.com/paritytech/web3-storage/pull/125#discussion_r3442998786_
>

### LINKED ISSUE #243: Migration: subxt::dynamic → typed storage_subxt bindings
Follow-up https://github.com/paritytech/web3-storage/pull/207#discussion_r3465733093

The `storage-subxt` crate (`storage-subxt/`) contains pre-generated, fully-typed Rust bindings for the parachain runtime, produced by `subxt-cli` from `metadata/storage_paseo_runtime.scale`. These bindings provide:

- **Typed transactions** via `runtime::tx().storage_provider().*()`, `runtime::tx().drive_registry().*()`, etc.
- **Typed storage queries** via `runtime::storage().storage_provider().*()`, etc.
- **Typed runtime API calls** via `runtime::apis().storage_provider_api().*()`.
- **Generated runtime types** under `runtime_types::*` — the canonical source of truth for all on-chain structs (e.g. `pallet_storage_provider::pallet::ProviderInfo`, `ProviderSettings`, `ProviderStats`, `BoundedVec`).

Before this work, several parts of the codebase used `subxt::dynamic::tx(...)` and `subxt::dynamic::storage(...)` with manually constructed `ScaleValue` trees, and declared local hand-rolled structs that duplicated fields already present in the generated types.

TODO:

Rust clients: `client/*`, `storage-interfaces/s3/client/src/substrate.rs`, `storage-interfaces/file-system/client/src/substrate.rs`
Use `storage-subxt` as trusted type definition

- [ ] Eliminate `subxt::dynamic` calls
- [ ] Replace hand-rolled types with generated equivalents
- [ ] Replace `sp-runtime` with `subxt_core` , `storage_subxt`

Note: The HTTP `/info` response from `provider-node` changed shape as a side effect of this migration. Any external HTTP client consuming `/info` must be updated.

danielbui12:
## Changes

- New crate `storage-subxt`: generated subxt typed bindings for the runtime, re-exporting subxt and subxt_signer as the single source of truth for those deps across the workspace
- storage-client
  - Replaced all `subxt::dynamic::tx/storage/constant` calls with typed `runtime::tx()` / `runtime::storage()` / `runtime::constants()` builders
  - Added `runtime_convert.rs`: byte-exact conversion layer between domain types (`AccountId32`, `H256`, `MultiSignature`, `AgreementTerms`, `Role`, `EndAction`, proof structs) and `generated runtime_types::*`
  - Replaced per-file manual `scale_value` decoders with direct typed-struct field access
  - Migrated `event_subscription.rs` to typed `as_event::<ev::*>()` decoding; deleted `scale_decode.rs`
  - Removed direct `subxt/subxt-signer` deps from `client/Cargo.toml`; all imports now flow through `storage_subxt::subxt` / `storage_subxt::subxt_signer`
- `provider-node`
  - `SubxtChainClient` rewritten: holds a `SubstrateClient` (from `storage_client`) instead of raw `OnlineClient` + `Keypair`; all coordinator trait impls use typed `storage::*` / `extrinsics::*`
  - `ChainMembershipResolver` in `auth.rs` replaced dynamic Buckets query + scale_value decoding with typed `storage::bucket_info()` + `rt::storage_primitives::Role` matching
  - Added three new functions to `client/src/substrate.rs`: `extrinsics::update_provider_multiaddr`, `extrinsics::provider_checkpoint`, `storage::checkpoint_config`

## Merge sequence
- 1. https://github.com/paritytech/web3-storage/pull/249
- 2. https://github.com/paritytech/web3-storage/pull/248
- 3. https://github.com/paritytech/web3-storage/pull/239

## Issue
- Relavant: https://github.com/paritytech/web3-storage/issues/64 
- Close https://github.com/paritytech/web3-storage/issues/243
- Close https://github.com/paritytech/web3-storage/issues/237


bkontur:
@danielbui12 please, make some TODO to check also using storage queries in TS SDK e.g.: https://github.com/paritytech/web3-storage/pull/160/changes/48d7560f2318f8bc5ba3a77cdcf003113c0365b0#diff-9c9ec00799f06105758439222211e7ef2fc782460416fb49c8234c3c1a8305efR174

danielbui12:
> @danielbui12 please, make some TODO to check also using storage queries in TS SDK e.g.: [48d7560#diff-9c9ec00799f06105758439222211e7ef2fc782460416fb49c8234c3c1a8305efR174](https://github.com/paritytech/web3-storage/commit/48d7560f2318f8bc5ba3a77cdcf003113c0365b0#diff-9c9ec00799f06105758439222211e7ef2fc782460416fb49c8234c3c1a8305efR174)

I created an issue https://github.com/paritytech/web3-storage/issues/243

pallet/src/lib.rs
```diff
@@ -3701,11 +3705,101 @@ pub mod pallet {
                     leaf_index: challenge.leaf_index,
                     chunk_index: challenge.chunk_index,
                     deadline: block.saturated_into::<u32>(),
+                    index: idx as u16,
                     deposit: challenge.deposit.saturated_into::<u128>(),
                 })
                 .collect()
         }
 
+        /// Query all challenges for a specific bucket.
```
bkontur: @danielbui12 ok, let's push this PR, I suggest couple of things:
- create `impls.rs` and move all this functions `impl<T: Config> Pallet<T> {` there
- create separate PR just for this changes touching `pallet` + `runtimes` (looks like just introducing new runtime apis)

And we continue with review there :)

pallet/src/runtime_api.rs
```diff
@@ -170,6 +173,15 @@ sp_api::decl_runtime_apis! {
...
 
+        /// Get all challenges for a specific bucket.
+        fn bucket_challenges(bucket_id: BucketId) -> Vec<ChallengeResponse>;
+
+        /// Get all challenges targeting a specific provider.
+        fn provider_challenges(provider: AccountId) -> Vec<ChallengeResponse>;
+
+        /// Get all challenges created by a specific challenger.
+        fn challenger_challenges(challenger: AccountId) -> Vec<ChallengeResponse>;
+
```
bkontur, line 184: @danielbui12 I would go with this naming (we can later align this for others in separate PR mentioned above):

```suggestion
        /// Get all challenges for a specific bucket.
        fn challenges_by_bucket_id(bucket_id: BucketId) -> Vec<ChallengeResponse>;

        /// Get all challenges targeting a specific provider.
        fn challenges_by_provider(provider: AccountId) -> Vec<ChallengeResponse>;

        /// Get all challenges created by a specific challenger.
        fn challenges_by_challenger(challenger: AccountId) -> Vec<ChallengeResponse>;

```

pallet/src/lib.rs
```diff
@@ -3701,11 +3705,101 @@ pub mod pallet {
...
+                                start_seq: challenge.start_seq,
+                                leaf_index: challenge.leaf_index,
+                                chunk_index: challenge.chunk_index,
+                                deadline,
+                                index: idx as u16,
+                                deposit: challenge.deposit.saturated_into::<u128>(),
+                            },
+                        )
+                })
+                .filter(|c| c.bucket_id == bucket_id)
```
bkontur: @danielbui12 looks like all the functions apply the same pattern:
- read the while storage list
- map all the items
- **filter**

we should filter ASAP and then map

pallet/src/lib.rs
```diff
@@ -3701,11 +3705,101 @@ pub mod pallet {
...
+            bucket_id: BucketId,
+        ) -> Vec<crate::runtime_api::ChallengeResponse> {
+            Challenges::<T>::iter()
+                .flat_map(|(block, challenges)| {
+                    let deadline: u32 = block.saturated_into::<u32>();
+                    challenges
+                        .into_iter()
+                        .enumerate()
+                        .map(
+                            move |(idx, challenge)| crate::runtime_api::ChallengeResponse {
```
bkontur: @danielbui12 looks like this mapping can be deduplicated (some function or maybe `From`?)

pallet/src/runtime_api.rs
```diff
@@ -132,11 +133,13 @@ pub struct ChallengeResponse {
     pub leaf_index: u64,
     pub chunk_index: u64,
     pub deadline: u32,
+    pub index: u16,
```
bkontur, line 137: @danielbui12 hmm, not sure if we need this `index` it is just random index which is generated from `enumerate` - I would remove that, but if we need this for some client processing, we should have some unique dedicated hash/id - for example if you delete challange (are they even  deleted?) the index will change or something. But let's continue with separate PR mentioned above for pallet/runtimes changes

pallet/src/lib.rs
danielbui12: @bkontur  I created a follow-up PR here https://github.com/paritytech/web3-storage/pull/256

pallet/src/runtime_api.rs
danielbui12:
```rust
    pub type Challenges<T: Config> =
        StorageMap<_, Blake2_128Concat, BlockNumberFor<T>, Vec<Challenge<T>>>;
```

concretely, Challenges is a mapping from block number to a vector of Challenge.
hence `index` is necessary to identify which Challenge are we want to interact with.

danielbui12: Update: https://github.com/paritytech/web3-storage/pull/125 already migrated Challanges to
```
  pub type Challenges<T: Config> = StorageDoubleMap<
        _,
        Blake2_128Concat,
        BlockNumberFor<T>,
        Twox64Concat,
        u16,
        Challenge<T>,
        OptionQuery,
    >;
```
now it is more intuitive that `index: u16` is essential

storage-subxt/Cargo.toml
```diff
@@ -0,0 +1,35 @@
...
+	"scale-info/std",
+	"serde/std",
+	"subxt-core/std",
+	"subxt-signer/std",
+	"subxt/jsonrpsee",
+	"subxt/native",
+]
+# Select the network runtime bindings exposed as `storage_runtime`.
+# Off (default): testnet (`storage_paseo_runtime`). On: mainnet (`storage_mainnet_runtime`).
+mainnet = []
```
ilchu, line 35: Considering we're nowhere close to having mainnet yet, you intend to just have it as a reminder here? As the paseo and localnet are virtually same.

storage-subxt/src/lib.rs
```diff
@@ -0,0 +1,37 @@
...
+pub use subxt_core;
+pub use subxt_signer;
+
+#[cfg(feature = "mainnet")]
+#[rustfmt::skip]
+pub mod storage_mainnet_runtime;
+#[cfg(feature = "mainnet")]
+pub use storage_mainnet_runtime::*;
+
+// #[cfg(not(feature = "mainnet"))]
```
ilchu, line 24:
```suggestion
#[cfg(not(feature = "mainnet"))]
```
or does it break something? Also why the need to skip formatter?

provider-node/src/subxt_client.rs
```diff
@@ -2,71 +2,66 @@
...
+/// Production implementation that talks to the chain via typed storage_client bindings.
 ///
-/// Cloning is cheap: `OnlineClient` shares one connection behind an `Arc`, and
+/// Cloning is cheap: `SubstrateClient` shares one connection behind an `Arc`, and
 /// the `Keypair` clone is a key copy. This lets a single instance be shared
 /// across every background coordinator.
 #[derive(Clone)]
 pub struct SubxtChainClient {
-    api: subxt::OnlineClient<subxt::PolkadotConfig>,
-    signer: subxt_signer::sr25519::Keypair,
```
ilchu, line 35: It's a nit, but when there's just one field and we're just making a thin wrapper, I'd rather have an unnamed struct.

provider-node/src/storage/disk.rs
```diff
@@ -8,13 +8,13 @@
 use super::NonceStore;
 use super::{BucketInfo, StorageBackend, StoredNode};
 use crate::error::Error;
-use crate::types::*;
 use codec::Encode;
 use rocksdb::{Options, DB};
-use sp_core::H256;
 use std::path::Path;
 use std::sync::{Arc, Mutex};
+use storage_client::provider_node_request_scheme::*;
```
bkontur: @danielbui12 ok, ok, I see, this PR tries to move "too much"/"lots of" stuff now. The problem why I created the https://github.com/paritytech/web3-storage/issues/178 is that at some point we started using storage-client as a dependency of provider-node, which is not the best thing, it should rather be the other way, the SDK client should use/include some shared deps from provider (e.g. subxt handling or common structs). Let me think, we need extract and clean carefully in the direction of https://github.com/paritytech/web3-storage/issues/178

provider-node/src/chain_state_coordinator.rs
```diff
@@ -457,19 +458,34 @@ mod tests {
...
-            max_capacity: 10_000,
-            min_duration: 10,
-            max_duration: 100,
-            price_per_byte: 5,
-            accepting_primary: true,
-            replica_sync_price: None,
-            accepting_extensions: true,
-            agreements_total: 3,
-            challenges_failed: 1,
+            settings: ProviderSettings {
```
bkontur, line 472: - [ ] @danielbui12 please, replicate/extract just this wrapping change here on the top of the `dev` branch and nothing else (+ looks like `packages` contains some related changes also)?
- [ ] plus you can also move those `#[api_version(2)]` and `pallet` changes

provider-node/src/auth.rs
```diff
@@ -144,100 +144,44 @@ impl ChainMembershipResolver {
...
-            }
-        }
+        let members = bucket
+            .members
+            .0
+            .into_iter()
+            .map(|m| {
+                let account = AccountId32::from(m.account.0);
+                use storage_subxt::api::runtime_types as rt;
+                let role = match m.role {
```
ilchu, line 172: Could reuse `to_role()` from `runtime_convert.rs` to not write all out here.

storage-subxt/src/lib.rs
```diff
@@ -0,0 +1,37 @@
...
+pub use storage_paseo_runtime::*;
+
+impl crate::api::runtime_types::pallet_storage_provider::runtime_api::AgreementResponse {
+    pub fn is_primary(&self) -> bool {
+        matches!(
+            self.role,
+            crate::api::runtime_types::storage_primitives::ProviderRole::Primary
+        )
+    }
+}
```
ilchu: I wouldn't put domain logic here, even though it's here because of orphan rule for codegen probably. Also I don't see it being used anywhere, so linter is not flagging due to the reexport?

client/src/admin.rs
```diff
@@ -451,11 +455,13 @@ impl AdminClient {
         nonce: u64, // nonce the providers signed over (echoed from their commitment)
         signatures: Vec<(String, Vec<u8>)>, // (provider SS58, signature bytes)
     ) -> ClientResult<()> {
+        // TODO: replace `nonce`, `signatures` with API call
+
```
ilchu, line 459: Do we have an issue for this? Or seems more or less within this PR's scope I'd say.

danielbui12: Yes, it is relevant to this https://github.com/paritytech/web3-storage/issues/64

storage-subxt/src/lib.rs
danielbui12: it breaks the clippy all features check

storage-subxt/Cargo.toml
danielbui12: I'm aiming to scaffold it as a complete crate.

client/src/admin.rs
ilchu: Well my only concern here is that we don't forget about this todo. #64 is a very broad umbrella issue and from the looks of it all the subissues are either closed or completed, so this one could be left unnoticed. But sure, I leave it to your judgment.

danielbui12: rust packages need to be re-org later but soon https://github.com/paritytech/web3-storage/pull/239#discussion_r3529542685

provider-node/src/auth.rs
danielbui12: runtime_convert convert from local types to subxt bindings.
this convert from subxt bindings back to local type

provider-node/src/chain_state_coordinator.rs
danielbui12: @bkontur 
- this changes come along with new subxt bindings, I dont get your point how to split them?
- "plus you can also move those #[api_version(2)] and pallet changes" not sure I get the new one. could you elaborate more?

provider-node/src/storage/disk.rs
danielbui12: I revert the provider-node migrations. keep only the subxt bidings-related changes

Issue #94, opened 2026-05-27
[Decentralisation] Protocol to interact with Provider Nodes directly


mudigal:
Currently we use REST APIs, this would be more centralized approach. We need to find a better decentralized way to write and read data.

Would Smoldot be good here or do we need to evaluate something else?

mudigal:
## Investigation Summary: Decentralized Read & Write Protocols

After analyzing the codebase and the current checkpoint/multi-provider architecture, here's a breakdown of the protocol options for replacing REST APIs with decentralized alternatives.

---

### The Core Asymmetry: Reads vs Writes

Reads and writes have fundamentally different properties and need different protocols:

- **Reads** are content-addressed — *any* peer with the data can serve it. This is naturally decentralized.
- **Writes** are inherently directed — they *must* reach the specific provider who holds the agreement, builds the MMR, and signs checkpoints. You cannot broadcast writes.

---

### Read Path: libp2p + Bitswap

Content-addressed reads map cleanly to existing P2P content distribution protocols:

| Layer | Current (REST) | Proposed |
|-------|---------------|----------|
| Discovery | RPC query for provider URL | DHT lookup by content hash (CID) |
| Transport | `GET /s3/:bucket_id/object` to one server | Bitswap request — any peer with the data serves it |
| Verification | Trust the provider | Content-addressed — hash verification on receipt |
| Availability | Single point of failure | Multiple peers (primaries + replicas) can serve |

Smoldot is relevant here for **light-client chain queries** (e.g., verifying which providers hold a bucket) without depending on a centralized RPC node.

---

### Write Path: Leader-Based Writes via libp2p

This is the harder problem. The current codebase reveals the answer through the checkpoint mechanism.

**Key constraint**: The `checkpoint` extrinsic (`pallet/src/lib.rs:2071`) requires all primary providers to sign the **same** `(mmr_root, start_seq, leaf_count)`. This means all primaries must converge on identical MMR state. If the client wrote to multiple primaries independently, each would assign different sequence numbers and produce divergent MMR roots — the checkpoint would fail.

#### Leader Election Is Deterministic and Off-Chain

The leader election does **NOT require any on-chain call**. The algorithm (`pallet/src/lib.rs:3654`) is pure deterministic math:

```rust
fn calculate_leader_index(bucket_id: BucketId, window: u64, num_providers: u32) -> u32 {
    let mut data = [0u8; 16];
    data[..8].copy_from_slice(&bucket_id.to_le_bytes());
[...]

PR #245, opened 2026-06-26
Remove disable auth flag

### LINKED ISSUE #232: Remove `disable_auth_i_know_what_i_am_doing` flag
Follow up to https://github.com/paritytech/web3-storage/pull/215, see this discussion: https://github.com/paritytech/web3-storage/pull/215#discussion_r3466228188

RafalMirowski1:
Removes the --disable-auth-i-know-what-i-am-doing flag.

closes https://github.com/paritytech/web3-storage/issues/232

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -242,7 +242,9 @@ jobs:
           done
 
           if [ -n "$TEST_FLAGS" ]; then
-            PROVIDER_COV=$(cargo llvm-cov $TEST_FLAGS \
+            # --lib counts the crate's own unit tests (matching the pallet job);
+            # $TEST_FLAGS adds the HTTP integration tests — measure both.
+            PROVIDER_COV=$(cargo llvm-cov --lib $TEST_FLAGS \
               -p storage-provider-node \
               --ignore-filename-regex='(\.cargo|rustc|_subxt\.rs|command\.rs|cli\.rs|main\.rs)' \
               --summary-only 2>&1 \
@@ -289,7 +291,9 @@ jobs:
           done
 
           if [ -n "$TEST_FLAGS" ]; then
-            BASE_PROVIDER_COV=$(cargo llvm-cov $TEST_FLAGS \
+            # --lib counts the crate's own unit tests (matching the pallet job);
+            # $TEST_FLAGS adds the HTTP integration tests — measure both.
+            BASE_PROVIDER_COV=$(cargo llvm-cov --lib $TEST_FLAGS \
               -p storage-provider-node \
               --ignore-filename-regex='(\.cargo|rustc|_subxt\.rs|command\.rs|cli\.rs|main\.rs)' \
               --summary-only 2>&1 \
--- a/client/examples/complete_workflow.rs
+++ b/client/examples/complete_workflow.rs
@@ -86,7 +86,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     println!("Establishing storage agreement on-chain...");
     let mut admin = AdminClient::new(chain_config.clone(), user_ss58.clone())?;
     admin.connect().await?;
-    admin.set_signer(user_keypair)?;
+    admin.set_signer(user_keypair.clone())?;
     let bucket_id = admin
         .establish_storage_agreement(provider_ss58, signed.terms, signed.signature)
         .await?;
@@ -99,7 +99,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
         provider_urls: vec![provider_url.to_string()],
         ..Default::default()
     };
-    let user = StorageUserClient::new(user_config)?;
+    let user = StorageUserClient::new(user_config)?.with_auth_signer(user_keypair);
     let data = b"hello e2e".to_vec();
     let data_root = user
         .upload(bucket_id, &data, ChunkingStrategy::default())
--- a/client/src/storage_user.rs
+++ b/client/src/storage_user.rs
@@ -14,13 +14,16 @@ use crate::encryption::{Cipher, EncryptionKey, XChaCha20Poly1305Cipher};
 use crate::verification::ClientVerifier;
 use base64::{engine::general_purpose::STANDARD as BASE64, Engine};
 use sp_core::H256;
-use storage_primitives::{blake2_256, BucketId};
+use storage_primitives::{blake2_256, build_auth_header, BucketId};
+use subxt_signer::sr25519::Keypair;
 
 /// Client for storage users (end users who store/retrieve data).
 pub struct StorageUserClient {
     base: BaseClient,
     verifier: ClientVerifier,
     cipher: Option<Box<dyn Cipher>>,
+    /// Keypair used to sign bucket-scoped provider requests.
+    auth_signer: Option<Keypair>,
 }
 
 impl StorageUserClient {
@@ -30,12 +33,49 @@ impl StorageUserClient {
             base: BaseClient::new(config)?,
             verifier: ClientVerifier::new(),
             cipher: None,
+            auth_signer: None,
         })
     }
 
     /// Create with default configuration.
     pub fn with_defaults() -> ClientResult<Self> {
         Self::new(ClientConfig::default())
+    }
+
+    /// Set the keypair that signs bucket-scoped provider requests.
+    pub fn with_auth_signer(mut self, signer: Keypair) -> Self {
+        self.auth_signer = Some(signer);
+        self
+    }
+
+    /// [`with_auth_signer`](Self::with_auth_signer) by mutable reference.
+    pub fn set_auth_signer(&mut self, signer: Keypair) {
+        self.auth_signer = Some(signer);
+    }
+
+    /// Attach the signed `Authorization` header (`method` = upper-case HTTP verb)
+    /// when an auth signer is configured; otherwise leave the request unchanged.
+    fn sign(
+        &self,
+        req: reqwest::RequestBuilder,
+        method: &str,
+        bucket_id: BucketId,
+    ) -> reqwest::RequestBuilder {
+        let Some(signer) = self.auth_signer.as_ref() else {
+            return req;
+        };
+        let timestamp = std::time::SystemTime::now()
+            .duration_since(std::time::UNIX_EPOCH)
+            .map(|d| d.as_secs())
+            .unwrap_or_default();
+        let header = build_auth_header(
+            &signer.public_key().0,
+            method,
+            bucket_id,
+            timestamp,
+            |msg| signer.sign(msg).0,
+        );
+        req.header("Authorization", header)
     }
 
     /// Enable client-side encryption with a custom cipher (builder pattern).
@@ -340,13 +380,12 @@ impl StorageUserClient {
                 .collect(),
         };
 
-        let response = self
+        let req = self
             .base
             .http
             .post(format!("{provider_url}/commit"))
-            .json(&request)
-            .send()
-            .await?;
+            .json(&request);
+        let response = self.sign(req, "POST", bucket_id).send().await?;
 
         if !response.status().is_success() {
             return Err(ClientError::Api(format!(
@@ -495,13 +534,12 @@ impl StorageUserClient {
             }),
         };
 
-        let response = self
+        let req = self
             .base
             .http
             .put(format!("{provider_url}/node"))
-            .json(&request)
-            .send()
-            .await?;
+            .json(&request);
+        let response = self.sign(req, "PUT", bucket_id).send().await?;
 
         if !response.status().is_success() {
             return Err(ClientError::Api(format!(
--- a/examples/papi/full-flow.ts
+++ b/examples/papi/full-flow.ts
@@ -90,9 +90,9 @@ async function setupAgreement(
   return bucketId;
 }
 
-async function uploadAndVerify(bucketId: bigint) {
+async function uploadAndVerify(bucketId: bigint, client: ChainSigner) {
   const payload = `Hello, Web3 Storage! [${new Date().toISOString()}] provider=${PROVIDER_SEED}`;
-  const { hash, data, commit } = await uploadChunk(PROVIDER_URL, bucketId, payload);
+  const { hash, data, commit } = await uploadChunk(PROVIDER_URL, bucketId, payload, client);
   console.log("  Uploaded %d bytes, mmr_root=%s", data.length, commit.mmr_root);
 
   const downloaded = await downloadChunk(PROVIDER_URL, hash);
@@ -176,7 +176,7 @@ async function main() {
     const bucketId = await setupAgreement(api, PROVIDER_URL, client, provider);
 
     console.log("\n=== Step 2: Upload data ===");
-    const upload = await uploadAndVerify(bucketId);
+    const upload = await uploadAndVerify(bucketId, client);
 
     console.log("\n=== Step 3: Off-chain challenge ===");
     const offchainId = await challengeOffchain(
--- a/examples/papi/full-flow.ts
+++ b/examples/papi/full-flow.ts
@@ -85,7 +85,11 @@ async function setupAgreement(
     signed.terms.valid_until
   );
   console.log("  Redeeming on-chain via establish_storage_agreement...");
-  const { bucketId } = await establishStorageAgreement(api, client, provider, signed);
+  // Finalize: the immediate upload reads bucket membership from the provider's
+  // finalized view, so an in-block establish would race it.
+  const { bucketId } = await establishStorageAgreement(api, client, provider, signed, {
+    mode: "finalized",
+  });
   console.log("  Bucket %s opened with primary agreement", bucketId);
   return bucketId;
 }
--- a/packages/layer1/src/fs/client.ts
+++ b/packages/layer1/src/fs/client.ts
@@ -5,8 +5,8 @@
  * surface. Chain ops delegate to the layer-0 pallet wrappers (silent, no
  * auto-retry, finalized submission + finalized reads by default — UI-grade,
  * reorg-safe; tests/examples opt into in-block/best via readOpts/submitMode);
- * HTTP ops go through core's retrying fetch and are signed when the signer
- * has a raw keypair.
+ * HTTP ops go through core's retrying fetch and are signed with the signer's
+ * raw keypair, which the provider always requires.
  *
  * Verification: `downloadByCid` is verified (single chunk — its hash IS the
  * CID; the layer-0 downloadChunk throws CidMismatchError). Path-based
--- a/primitives/src/lib.rs
+++ b/primitives/src/lib.rs
@@ -405,6 +405,46 @@ pub fn hash_children(left: H256, right: H256) -> H256 {
 }
 
 // ─────────────────────────────────────────────────────────────────────────────
+// Provider HTTP authentication
+// ─────────────────────────────────────────────────────────────────────────────
+
+fn hex_lower(bytes: &[u8]) -> alloc::string::String {
+    use core::fmt::Write;
+    let mut s = alloc::string::String::with_capacity(bytes.len() * 2);
+    for b in bytes {
+        let _ = write!(s, "{b:02x}");
+    }
+    s
+}
+
+/// The canonical message a client signs for a bucket-scoped request:
+/// `web3storage:<METHOD>:<bucket_id>:<timestamp>` (`METHOD` upper-case, `timestamp`
+/// Unix seconds). The provider rebuilds this exact string to verify the signature.
+pub fn auth_message(method: &str, bucket_id: BucketId, timestamp: &str) -> alloc::string::String {
+    alloc::format!("web3storage:{method}:{bucket_id}:{timestamp}")
+}
+
+/// Build the provider's `Authorization` header value: the sr25519 signature of
+/// [`auth_message`] formatted as `Web3Storage <pubkey_hex>:<signature_hex>:<timestamp>`.
+/// `sign` returns the 64-byte signature, keeping this keypair-type agnostic.
+pub fn build_auth_header(
+    pubkey: &[u8; 32],
+    method: &str,
+    bucket_id: BucketId,
+    timestamp: u64,
+    sign: impl FnOnce(&[u8]) -> [u8; 64],
+) -> alloc::string::String {
+    let timestamp = alloc::format!("{timestamp}");
+    let signature = sign(auth_message(method, bucket_id, &timestamp).as_bytes());
+    alloc::format!(
+        "Web3Storage 0x{}:0x{}:{}",
+        hex_lower(pubkey),
+        hex_lower(&signature),
+        timestamp
+    )
+}
+
+// ─────────────────────────────────────────────────────────────────────────────
 // Verification Utilities
 // ─────────────────────────────────────────────────────────────────────────────
 
--- a/provider-node/src/auth.rs
+++ b/provider-node/src/auth.rs
@@ -188,64 +188,108 @@ impl MembershipResolver for ChainMembershipResolver {
             None => return Ok(vec![]),
         };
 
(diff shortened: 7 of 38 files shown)


provider-node/src/auth.rs
```diff
@@ -177,64 +188,108 @@ impl MembershipResolver for ChainMembershipResolver {
             None => return Ok(vec![]),
         };
```
RafalMirowski1, line 188: Decoding was bugged after enabling the auth, here is some ad-hoc fix but the proper way is to wait for https://github.com/paritytech/web3-storage/pull/239

.github/workflows/check.yml
```diff
@@ -242,7 +242,9 @@ jobs:
           done
 
           if [ -n "$TEST_FLAGS" ]; then
-            PROVIDER_COV=$(cargo llvm-cov $TEST_FLAGS \
+            # --lib counts the crate's own unit tests (matching the pallet job);
+            # $TEST_FLAGS adds the HTTP integration tests — measure both.
+            PROVIDER_COV=$(cargo llvm-cov --lib $TEST_FLAGS \
```
bkontur: @RafalMirowski1 what is the consequence of this? Unit-tests are also accounted? So, better overall coverage score/number with `--lib` for provider?

Relates to: https://github.com/paritytech/web3-storage/pull/196

RafalMirowski1: yes, with --lib unit tests are accounted, so the reported coverage number is higher. After I added this
provider coverage on this PR went from 59.41% to 69.43%. I think if something is covered by unit tests we should count it or do we only want to count integration tests?
EDIT:
I guess here https://github.com/paritytech/web3-storage/pull/196#discussion_r3421157638 Naren says not including units was intentional.

bkontur: > yes, with --lib unit tests are accounted, so the reported coverage number is higher. After I added this provider coverage on this PR went from 59.41% to 69.43%. I think if something is covered by unit tests we should count it or do we only want to count integration tests? EDIT: I guess here [#196 (comment)](https://github.com/paritytech/web3-storage/pull/196#discussion_r3421157638) Naren says not including units was intentional.

right, I see, can you please, extract this change from this PR? We can merge this change with https://github.com/paritytech/web3-storage/pull/196 (let's think about the coverage usage there)

storage-interfaces/s3/client/src/lib.rs
```diff
@@ -145,13 +145,18 @@ impl S3Client {
...
             .map_err(|e| S3ClientError::ProviderError(e.to_string()))?;
 
         let substrate_client = SubstrateClient::new(chain_url, seed_phrase)
             .await
             .map_err(|e| S3ClientError::ChainError(e.to_string()))?;
 
+        // Reuse the same key to authenticate bucket-scoped provider requests.
+        if let Ok(keypair) = substrate_client.signer_keypair() {
+            storage_client.set_auth_signer(keypair);
+        }
```
ilchu: Here there could be danger of fallthrough, but by virtue of construction this should never fail. So my suggestion would be to unwrap the keypair from option in S3 layer as well.

storage-interfaces/s3/client/src/substrate.rs
```diff
@@ -96,6 +96,15 @@ impl SubstrateClient {
             .ok_or_else(|| "No signer configured".to_string())
     }
 
+    /// Get the signer keypair (cloned), for reuse by other components such as
+    /// the Layer 0 client's provider-request authentication.
+    pub fn signer_keypair(&self) -> std::result::Result<Keypair, String> {
```
ilchu: This is pretty much a duplicate of `.signer()` right above it (barring a reference).

client/src/storage_user.rs
```diff
@@ -38,6 +42,42 @@ impl StorageUserClient {
...
+    /// when an auth signer is configured; otherwise leave the request unchanged.
+    fn sign(
+        &self,
+        req: reqwest::RequestBuilder,
+        method: &str,
+        bucket_id: BucketId,
+    ) -> reqwest::RequestBuilder {
+        let Some(signer) = self.auth_signer.as_ref() else {
+            return req;
+        };
```
ilchu: If we have this short circuit then we can't really say that auth route is enforced, or am I missing something here? Like we should force the presence of a keypair at all times too, otherwise no actual signing and authorization is happening.

client/src/storage_user.rs
```diff
@@ -14,13 +14,16 @@ use crate::encryption::{Cipher, EncryptionKey, XChaCha20Poly1305Cipher};
...
+use storage_primitives::{blake2_256, build_auth_header, BucketId};
+use subxt_signer::sr25519::Keypair;
 
 /// Client for storage users (end users who store/retrieve data).
 pub struct StorageUserClient {
     base: BaseClient,
     verifier: ClientVerifier,
     cipher: Option<Box<dyn Cipher>>,
+    /// Keypair used to sign bucket-scoped provider requests.
+    auth_signer: Option<Keypair>,
```
ilchu: I believe we'd benefit from making it mandatory.

provider-node/tests/common/mod.rs
```diff
@@ -0,0 +1,157 @@
+// SPDX-License-Identifier: GPL-3.0-only
+
+//! Shared test helpers for the provider-node integration suites.
+//!
+//! Tests run with `//Alice` as a bucket `Admin`
+//! ([`with_admin_member`]) and use [`SignedClient`] to sign every request as
+//! `//Alice`.
+
+#![allow(dead_code)]
```
ilchu, line 9:
```suggestion

```

client/src/storage_user.rs
bkontur: > If we have this short circuit then we can't really say that auth route is enforced, or am I missing something here? Like we should force the presence of a keypair at all times too, otherwise no actual signing and authorization is happening.

Just thinking (without looking in to code and thinking about that much), maybe part of the client is readonly and it does not require any signing, but probably in this case there shouldn't be any `Option` also. But I agree with @ilchu here, `sign` function by-passing the actual signing with `auth_signer: Option<...>` does look like a good pattern, I would rather throw error instead. But also need to check the whole PR and I am also not that familar with this client code.

user-interfaces/drive-ui/src/lib/drive-client.ts
```diff
@@ -120,14 +122,15 @@ export class DriveClient {
     }
     let chainSigner: ChainSigner | null = null;
     if (this.signer && this.signerAddress) {
-      // Wallet flows hand us a PolkadotSigner + address; recover the public
-      // key from the address. No raw keypair here, so provider requests go
-      // unsigned (same as this app always behaved).
-      const [publicKey] = ss58Decode(this.signerAddress);
+      // drive-ui derives raw dev-account keypairs, so provider requests are
```
bkontur: why is this mentioning `dev-account`. The UIs should be account-agnostic, no UI code should directly work with dev-accounts. I know that we support that, but it should be "one-place" where we either allow using dev-accounts or not (per network configuration). I mean the UIs code should not think about "what kind of account" is this and do some IF-ing (ok, as I said before unless really necessary handling cfg). Other words, if the UI allows to select dev-account, ok, no problem, but the rest of the code should handle it as a regular account (signing and everything), no distinction dev vs non-dev account. 

@RafalMirowski1 I know this is not related to your changes, but please, we need to clean this also for UIs/clients/SDKs - no explicit usage of dev-accounts. Even selected dev-account should be converted to `ChainSigner` or whatever without any assumptions or exceptions. I am not sure know, if this could/should land as a separate PR before this one, or we can join it here? What is easier for you :)

.github/workflows/check.yml
RafalMirowski1: ok, removed this

storage-interfaces/s3/client/src/substrate.rs
RafalMirowski1: removed

client/src/storage_user.rs
RafalMirowski1: Now `signer` is required so removed this short circuit. Also I created a `Signer` new type which wraps `Arc<Keypair>` internally, because we had duplication like `set_signer` + `set_dev_signer` - not sure if we need this, but made sense to me

client/src/storage_user.rs
RafalMirowski1: no longer an `Option`

storage-interfaces/s3/client/src/lib.rs
RafalMirowski1: removed, I pass `signer` directly into constructor now.

user-interfaces/drive-ui/src/lib/drive-client.ts
RafalMirowski1: Some loose thoughts here: agree we should have uniform `ChainSigner`. I think also above the problem is that wallets sign like `<Bytes>{msg}</Bytes>` and don't expose the raw keypair to us. Before auth was not enforced so it was fine for UI to fallback to unsigned requests but now that would be broken, so when we do verify_signature we should probably make a change there to accept wallet-style signing and then also come up with a way not to have bad UX and pop-ups all the time. We then could drop keypair and `DevAccountName` from `ChainSigner`. Seems large enough in scope to do as a follow-up on top of this one, and I'd move the rest of the dev-account cleanup there too. What do you think?

primitives/src/lib.rs
```diff
@@ -593,6 +593,46 @@ pub fn hash_children(left: H256, right: H256) -> H256 {
     blake2_256(&data)
 }
 
+// ─────────────────────────────────────────────────────────────────────────────
+// Provider HTTP authentication
+// ─────────────────────────────────────────────────────────────────────────────
+
+fn hex_lower(bytes: &[u8]) -> alloc::string::String {
```
bkontur: @RafalMirowski1 please, move this hex_lower/auth_message/build_auth_header to the provider-node file, this is provider code, and primitives is supposed to be common module for sharing some common code between pallets and provider code - #178

client/examples/complete_workflow.rs
```diff
@@ -86,7 +86,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     println!("Establishing storage agreement on-chain...");
     let mut admin = AdminClient::new(chain_config.clone(), user_ss58.clone())?;
     admin.connect().await?;
-    admin.set_signer(user_keypair)?;
+    admin.set_signer(user_keypair.clone().into())?;
```
bkontur: @RafalMirowski1 this `clone` caught my eyes, why we use the same key/secret for admin and the user? Shouldn't it to be two different accounts?

client/examples/complete_workflow.rs
bkontur:
```suggestion
    let user_client_config = ClientConfig {
```

client/examples/complete_workflow.rs
bkontur: can we pass it to `AdminClient::new(` like for `StorageUserClient`?
```suggestion
```

client/src/signer.rs
```diff
@@ -0,0 +1,114 @@
...
+    pub fn from_seed(seed: &str) -> ClientResult<Self> {
+        let uri = SecretUri::from_str(seed)
+            .map_err(|e| ClientError::Config(format!("invalid signer seed: {e}")))?;
+        Keypair::from_uri(&uri)
+            .map(|keypair| Self(Arc::new(keypair)))
+            .map_err(|e| ClientError::Config(format!("invalid signer seed: {e}")))
+    }
+
+    /// A well-known dev account by name: `"alice"`..`"ferdie"` (case-insensitive).
+    pub fn dev(name: &str) -> ClientResult<Self> {
```
bkontur: not sure where this is used, but probably should be feature-guarded just for tests maybe?

client/src/signer.rs
```diff
@@ -0,0 +1,114 @@
...
+        let keypair = match name.to_ascii_lowercase().as_str() {
+            "alice" => dev::alice(),
```
bkontur: well, usually we use `sp_keyring` for this dev accounts

examples/papi/full-flow.ts
```diff
@@ -85,15 +88,19 @@ async function setupAgreement(
...
+  // finalized view, so an in-block establish would race it.
+  const { bucketId } = await establishStorageAgreement(api, client, provider, signed, {
+    mode: "finalized",
+  });
   console.log("  Bucket %s opened with primary agreement", bucketId);
   return bucketId;
 }
 
-async function uploadAndVerify(api: ParachainApi, bucketId: bigint) {
+async function uploadAndVerify(api: ParachainApi, bucketId: bigint, client: ChainSigner) {
```
bkontur: why `client` instead of `signer`?
```suggestion
async function uploadAndVerify(api: ParachainApi, bucketId: bigint, signer: ChainSigner) {
```

packages/layer1/src/fs/client.ts
```diff
@@ -120,8 +120,7 @@ export class FileSystemClient {
   }
 
   private authHeaders(method: string, bucketId: bigint): Record<string, string> {
```
bkontur: @RafalMirowski1 can we deduplicate this `authHeaders` function (+ probably more, signProviderRequest, ...) between `fs` and `s3`? E.g. move to `layer0` or `base`?


Image

storage-interfaces/file-system/client/examples/ci_integration_test.rs
```diff
@@ -65,10 +65,8 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
 
     // Step 1: Create the client
     println!("Step 1: Creating file system client...");
-    let mut fs_client = FileSystemClient::new(chain_ws, provider_url)
-        .await?
-        .with_dev_signer("alice")
-        .await?;
+    let mut fs_client =
+        FileSystemClient::new(chain_ws, provider_url, Signer::dev("alice")?).await?;
```
bkontur: 👍

storage-interfaces/file-system/client/src/substrate.rs
```diff
@@ -23,72 +21,32 @@ pub const PALLET_NAME: &str = "DriveRegistry";
 #[derive(Clone)]
 pub struct SubstrateClient {
```
bkontur, line 22: @RafalMirowski1 I didn't check in deep, but looks like we have multiple `pub struct SubstrateClient {` which looks pretty same, I think that we should simplify this and have just one in `client`, probably they are doing the same thing - please, let's create and follow-up issue for this clients/SDKs cleanup and do it when #178 lands (link it here please, as an follow-up)

user-interfaces/drive-ui/src/lib/drive-client.ts
```diff
@@ -146,6 +144,11 @@ export class DriveClient {
     this.rebuild();
   }
 
+  setKeypair(keypair: Keypair | null): void {
```
bkontur: @RafalMirowski1 I am not that node guy, but having both/separate `setSigner` and `setKeypair` seems strange, I would expect just one think `setUser` or `setAccount` or `setSigner` which resets everything when we switch accounts in UI, can you please, check this?

user-interfaces/drive-ui/src/state/drive.state.ts
```diff
@@ -93,6 +93,14 @@ combineLatest([signer$$, signerAddress$$]).subscribe(([signer, address]) => {
   client.setSigner(signer, address);
 });
 
+// The raw keypair signs provider HTTP requests (/fs uploads, listing, delete).
+// Own subscription, mirroring setSigner above; rebuild() folds it into the
+// ChainSigner. A missing keypair means unsigned requests, which the provider
+// rejects with 401.
+keypair$$.subscribe((keypair) => {
```
bkontur: relates to https://github.com/paritytech/web3-storage/pull/245/changes#r3579022987

provider-node/src/command.rs
```diff
@@ -137,31 +137,18 @@ pub async fn run() -> Result<(), Box<dyn std::error::Error>> {
 }
 
 /// Apply CLI-derived CORS and auth settings to a freshly constructed state.
-///
-/// Authentication is enforced by default: unless the operator explicitly passed
-/// `--disable-auth-i-know-what-i-am-doing`, we wire in the on-chain membership
-/// resolver so every bucket-scoped request is checked against the caller's role.
 fn configure_state(state: ProviderState, cli: &Cli) -> ProviderState {
```
bkontur: @RafalMirowski1 I would like to clean up this provider configuration little bit now, something like this, because we have here multiple patterns -> take state: ProviderState -> return new state, lots of this mutable setter `set_auth_config` ...

```                                                                                                                                                                                                                        
/// Everything a servable ProviderState requires. No Options.                                                                                                                                                           
  pub struct ProviderDeps {                                                                                                                                                                                               
      pub storage: Arc<dyn StorageBackend>,                                                                                                                                                                               
      pub nonce_store: Arc<dyn NonceStore>,                                                                                                                                                                               
      pub membership: Arc<auth::MembershipCache>,                                                                                                                                                                         
      pub auth_max_skew: Duration,                                                                                                                                                                                        
  }                                                                                                                                                                                                                       
                                                                                                                                                                                                                          
  impl ProviderState {                                                                                                                                                                                                    
      pub fn with_seed(deps: ProviderDeps, seed: &str) -> Result<Self, String> { ... }                                                                                                                                    
      pub fn with_provider_id(deps: ProviderDeps, provider_id: String) -> Self { ... }                                                                                                                                    
      // CORS is the only genuine optional → stays a consuming builder:                                                                                                                                                   
      pub fn with_cors_origins(mut self, origins: Option<Vec<String>>) -> Self { ... }                                                                                                                                    
  }                                                                                                                                                                                                                       
```                                                                                                                                                                                                                          

                                                                                                                                                                                                                          
  This deletes:                                                                                                                                                                                                           
  - membership_cache: Option<...> → plain Arc<MembershipCache>                                                                                                                                                            
  - set_auth_config — constructor arg now                                                                                                                                                                                 
  - the "No membership cache" 500 branch — unrepresentable                                                                                                                                                                
  - set_nonce_store + its Arc::get_mut fallback — the "must be called while solely owned" doc contract becomes the type signature                                                                                         
  - configure_state — nothing left to configure post-hoc

client/examples/complete_workflow.rs
danielbui12: I did that :D 
Basically, this file walks through a happy flow:
- Create a bucket
- Upload object
- Download data then verify data integrity

The rust client is split into 2 components:
- `AdminClient`: contains `establish_storage_agreement` (creating a bucket).
- `StorageUserClient`: provides the object upload functionality.

I don't think using the same key for this flow is problematic.
If we use 2 different keys, we would need an additional step to invoke `set_member` to grant read/write permissions before uploading objects.

both work!

primitives/src/lib.rs
```diff
@@ -593,6 +593,46 @@ pub fn hash_children(left: H256, right: H256) -> H256 {
...
+    s
+}
+
+/// The canonical message a client signs for a bucket-scoped request:
+/// `web3storage:<METHOD>:<bucket_id>:<timestamp>` (`METHOD` upper-case, `timestamp`
+/// Unix seconds). The provider rebuilds this exact string to verify the signature.
+pub fn auth_message(method: &str, bucket_id: BucketId, timestamp: &str) -> alloc::string::String {
+    alloc::format!("web3storage:{method}:{bucket_id}:{timestamp}")
+}
+
```
danielbui12: Replay attack vulnerability.
The message binds none of path, body, provider, nonce. 
Attacker can capture header of a bucket then put arbitrary bodies on any requests for that bucket, and replays across providers

client/src/storage_user.rs
```diff
@@ -501,13 +528,12 @@ impl StorageUserClient {
...
-        let response = self
+        let req = self
             .base
             .http
             .put(format!("{provider_url}/node"))
-            .json(&request)
-            .send()
-            .await?;
+            .json(&request);
+        let response = self.sign(req, "PUT", bucket_id).send().await?;
```
danielbui12, line 532: https://github.com/paritytech/web3-storage/pull/245#discussion_r3585550318

client/src/storage_user.rs
```diff
@@ -342,13 +370,12 @@ impl StorageUserClient {
...
-        let response = self
+        let req = self
             .base
             .http
             .post(format!("{provider_url}/commit"))
-            .json(&request)
-            .send()
-            .await?;
+            .json(&request);
+        let response = self.sign(req, "POST", bucket_id).send().await?;
```
danielbui12, line 374: https://github.com/paritytech/web3-storage/pull/245#discussion_r3585550318

primitives/src/lib.rs
danielbui12: @RafalMirowski1  since we already have RBAC in pallet storage provider, I wonder why do we need this?

primitives/src/lib.rs
RafalMirowski1: moved to negotiation - feels a bit weird place but used by both client and provider to have same source of `auth_message`

packages/layer1/src/fs/client.ts
RafalMirowski1: extracted a common base client in `base-client.ts`

provider-node/src/command.rs
RafalMirowski1: `ProviderDeps` is now as above.

primitives/src/lib.rs
RafalMirowski1: > @RafalMirowski1 since we already have RBAC in pallet storage provider, I wonder why do we need this?

This is for http endpoints, so I think signature is used to prove to the provider that you are who you say you are, otherwise if say Alice is Writer and provider http endpoints are open, I can claim I'm Alice. Or am i misunderstanding this?



> Replay attack vulnerability. The message binds none of path, body, provider, nonce. Attacker can capture header of a bucket then put arbitrary bodies on any requests for that bucket, and replays across providers

Yes, but I did not change it here since I think we're going to have to fix signature anyway to enable wallet signing: https://github.com/paritytech/web3-storage/pull/245#discussion_r3549962873 since enabling auth breaks it with current scheme.

danielbui12: > Yes, but I did not change it here since I think we're going to have to fix signature anyway to enable wallet signing: https://github.com/paritytech/web3-storage/pull/245#discussion_r3549962873 since enabling auth breaks it with current scheme.

I know, we may also change from HTTP to P2P in the future https://github.com/paritytech/web3-storage/issues/94#issuecomment-4777560110

I'm wondering should we keep insecure code? wdyt @bkontur 

>  since enabling auth breaks it with current scheme.

@RafalMirowski1  could you elaborate more? I did not get this

RafalMirowski1: sure, i wasn't being clear, I meant the provider verifies a raw sr25519 signature over web3storage:<METHOD>:<bucket_id>:<timestamp>, but wallet extensions only sign the `<Bytes>`-wrapped form and don't expose the raw key so with changes here, any wallet extension flow in our UIs will fail (before there wouldve been a fallback to unsigned which worked with auth disabled)

danielbui12: `<Bytes>` wrapper is due to sign function, I also faced this weird problem before.
But my finding is about message scheme.

> The message binds none of path, body, provider, nonce.
> Attacker can capture header of a bucket then put arbitrary bodies on any requests for that bucket, and replays across providers

Here is my suggestion. Adding explicit provider account and http request body to auth message scheme.
```suggestion
pub fn auth_message(method: &str, bucket_id: BucketId, timestamp: &str) -> alloc::string::String {
    alloc::format!("web3storage:{method}:{provider_acc}:{bucket_id}:{explicit_http_request_body}:{timestamp}")
}
    
```

bkontur: I am also not that sure about this wrapping stuff, should be reflected by design, I created separate issue for this to figure out this auth scheme: https://github.com/paritytech/web3-storage/issues/304 (for you both :) )

PR #252, opened 2026-06-30
Encapsulate `Commitment` struct, group extrinsic args

### LINKED ISSUE #250: Encapsulate the `Commitment` structure
Currently quite a few pallet calls have a `too_many_arguments` lint disabled. They get passed separate fields of what could be called `Commitment` or `CommitmentRange` together with e.g. `BucketLocation`.

First discussed in:
https://github.com/paritytech/web3-storage/pull/125#discussion_r3490945121

ilchu:
Follow-up to #125, addressing bkontur's review note on the duplicated `(mmr_root, start_seq, leaf_count)` triplet and the `#[allow(clippy::too_many_arguments)]` on the challenge/checkpoint extrinsics.

Closes #250.

## What

Introduces two shared primitives — `Commitment { mmr_root, start_seq, leaf_count }` (with `range_end()`/`contains_seq()` defined once) and `ChunkLocation { leaf_index, chunk_index }` — and uses them two ways:

- **Dedup.** Embeds `Commitment` in `CommitmentPayload`, `BucketSnapshot`, `ReplicaSyncRecord`, and `CheckpointProposal` in place of the loose fields. SCALE encodes a nested struct as the bare concatenation of its fields and the triplet was contiguous and in the same order everywhere, so the byte layout is unchanged — signed payloads still verify and stored values need no migration (a byte-layout test pins this).
- **Grouped extrinsic args.** `checkpoint`, `provider_checkpoint`, `challenge_offchain`, `challenge_checkpoint`, and `challenge_replica` now take `commitment` / `location` instead of loose `u64`s. The bodies destructure at the top, so the logic is unchanged; `challenge_offchain` drops to 7 args and `checkpoint` to 4, so both `too_many_arguments` allows are removed. The stored `Challenge` folds `leaf_index`/`chunk_index` into `location: ChunkLocation` (byte layout unchanged).

`leaf_index` stays distinct from `leaf_count` throughout — it's a position within the MMR, not part of the commitment range — so the two are never conflated.

## Callers

Updated across the stack: the storage-provider precompile, the Rust client dynamic builders (with `nonce`/`leaf_count` threading reconciled), provider-node's `provider_checkpoint` call, the regenerated PAPI descriptors in `packages/papi/.papi`, and the layer0 TS tx builders + the e2e examples.


pallet/src/lib.rs
bkontur: @ilchu what about here pass `commitment`?
```suggestion
                    CommitmentPayload::new(bucket_id, commitment, nonce);
```

Looks like it could simplify few lines bellow:

Image

plus:
```
- Self::update_historical_roots(bucket, current_block, mmr_root);
+ Self::update_historical_roots(bucket, current_block, payload.mmr_root);
```


Also thinking about changing CommitmentPayload+BucketCheckpointed itself to use Commitment wrapper inside directly (+ change design)

pallet/src/tests/challenge.rs
bkontur, line 2133: Do we have more places like this? That can be wrapped?

```diff
-        mmr_root: H256,
-        start_seq: u64,
-        leaf_count: u64,
+        commitment: Commitment
```

docs/design/scalable-web3-storage-implementation.md
```diff
@@ -676,9 +676,7 @@ pub enum Event<T: Config> {
     },
     BucketCheckpointed {
         bucket_id: BucketId,
-        mmr_root: H256,
-        start_seq: u64,
-        leaf_count: u64,
+        commitment: Commitment,
```
bkontur, line 680: @eskimor are you ok with this [`Commitment` wrapper](https://github.com/paritytech/web3-storage/pull/252/changes#diff-93cb951626774323817cabfcdf4c2db9940f66818168e287b6cb9fd725dbb0b0R320) for `(mmr_root, start_seq, leaf_count)` - it simplifies and deduplicates code on the way little bit

@ilchu if ok, we should also define/introduce this `struct Commitment` in this implementation.md

eskimor: Sure!

Issue #101, opened 2026-05-27
Evaluate database engine for Storage Provider and Blockchain Provider


mudigal:
## Context

The Storage Provider and Blockchain Provider have different database workload profiles but both need production-grade database selection and tuning. This issue tracks the evaluation of database engines and configuration for both components.

Related: Issue #100 (per-bucket database isolation) — the engine choice here directly informs the per-bucket architecture.

## Current State

- **Storage Provider**: RocksDB 0.22 with 3 column families (`CF_NODES`, `CF_BUCKETS`, `CF_ROOT_TO_BUCKET`). See `provider-node/src/storage/disk.rs`.
- **Blockchain Provider**: Uses Substrate's default backend (RocksDB via `sc-client-db`).

## Database Trade-Off Matrix

| Architectural Vector | RocksDB (LSM-Tree) | ParityDB (Hash-Based) | Sled (B-Tree) | SQLite (WAL) |
|---------------------|--------------------|-----------------------|---------------|--------------|
| Primary Structure | Log-Structured Merge-Trees (SSTs) | Fixed-size value tables with hash-indexed mmap | Lock-free B-Tree | B-Tree with WAL journaling |
| Ideal Workloads | High-volume sequential reads/writes | High-frequency state trie lookups (32-byte keys) | Mixed read/write with transactional safety | Single-writer, many-reader with complex queries |
| Write Amplification | High (compaction rewrites data across levels) | Low (append-only) | Low-moderate | Low (WAL append) |
| Memory Management | Explicit app-level block caches | Implicit (OS page cache) | Implicit (OS page cache) | Configurable page cache |
| Rust Native | No (FFI to C++) | Yes | Yes | No (FFI to C) |
| Deletion Cost | High (tombstones trigger compaction) | Moderate | Moderate | Low (single file delete for per-bucket) |
| Maturity | Very high (Facebook, production-proven) | Moderate (Parity-maintained) | Moderate (community) | Very high (ubiquitous) |

### RocksDB Bottlenecks
- **SSD Compaction Wear**: Continuous background compaction generates intense disk writes, reducing SSD lifespans and competing for write bandwidth.
- **FFI Context Switching**: Interacting via FFI introduces CPU overhead noticeable during high-frequency operations.
- **Deletion Latency Spikes**: Clearing expired agreements writes massive volumes of tombstones, triggering compactions that can delay other operations.
[...]

Issue #100, opened 2026-05-27
Evaluate per-bucket database isolation on Storage Provider


mudigal:
## Problem Statement

The current Storage Provider uses a single RocksDB instance with 3 column families (`CF_NODES`, `CF_BUCKETS`, `CF_ROOT_TO_BUCKET`) for all buckets. This creates a **structural scaling bottleneck** driven primarily by how MMR state is stored and rebuilt per bucket.

### The Core Problem: MMR Rebuild Is O(n) Per Operation

`BucketState` stores all MMR leaves as a `Vec<MmrLeaf>` serialized into a single value in `CF_BUCKETS`. Every `commit()` and `get_mmr_proof()` call:

1. Deserializes the entire bucket state (all leaves) from one RocksDB key
2. **Rebuilds the entire MMR from scratch** by replaying all leaves
3. Appends new leaves (for commits)
4. Reserializes everything back to disk

```rust
// This runs on EVERY commit and EVERY proof generation (disk.rs:351-356, 436-438)
let mut mmr = crate::mmr::Mmr::new();
for leaf in &bucket.leaves {
    mmr.push(blake2_256(&leaf.encode()));
}
```

For a bucket with 100,000 leaves, every single upload replays 100,000 hashes. The serialized `BucketState` blob grows linearly and unboundedly — at scale this becomes a multi-MB single-key value that gets deserialized/reserialized on every operation.

### Secondary Problems

**Cross-bucket contention on the write path**: All bucket commits go through the same RocksDB write buffer. A burst of commits to Bucket A blocks Bucket B's checkpoint signature response, which has a 48-hour challenge deadline.

**Challenge response latency**: `get_mmr_proof()` also rebuilds the entire MMR just to generate one proof. Under load, generating proofs for large buckets competes with all other bucket operations for the same DB resources.

**No tenant isolation**: Storage and query spikes from one client's bucket affect all other buckets on the same provider.

**Expensive bucket deletion**: Deleting an expired bucket requires writing individual tombstone markers for every key, triggering RocksDB compaction cycles that compete with active storage operations.

**Migration difficulty**: Moving a single bucket to another provider requires scanning and extracting its data from the shared database — there's no self-contained unit to transfer.

### Architecture

Each bucket gets its own isolated database instance alongside a node-wide global index:

```
Storage Provider
├── Global Index (single lightweight DB)
│   ├── Bucket registry (bucket_id → DB path, status, stats)
│   ├── Deduplication index (content_hash → reference count)
[...]

PR #253, opened 2026-06-30
Database research

franciscoaguirre:
Addresses #101 and #100

Benchmarked different database options according to #101 and wrote some results to come to a decision.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -643,6 +643,7 @@ For the full review criteria (Parity Standards), see the `/review` skill. The re
 | [Storage Marketplace](docs/design/marketplace.md) | Provider capacity & discovery |
 | [Checkpoint Protocol](docs/design/CHECKPOINT_PROTOCOL.md) | Automated checkpoint management |
 | [File System Architecture](docs/filesystems/ARCHITECTURE.md) | Layer 1 encoding, security, blockchain details |
+| [Database Engine Evaluation](docs/design/database-evaluation/README.md) | Benchmark-backed DB selection |
 
 ## Common Issues & Solutions
 
--- a/benchmarks/db-bench/src/workloads/state_trie.rs
+++ b/benchmarks/db-bench/src/workloads/state_trie.rs
@@ -1,0 +1,222 @@
+//! Blockchain Node workloads — a synthetic model of Substrate's state-trie
+//! access pattern over the raw engine: random 32-byte keys, batched commits per
+//! "block", random point reads, and pruning of old keys.
+//!
+//! This is deliberately a *microbenchmark on the raw key/value engine*, not a
+//! full Substrate node. It isolates the engine's behaviour under the trie
+//! access pattern; the faithful end-to-end check (omni-node `--database
+//! rocksdb` vs `--database paritydb`) is documented in the report as follow-up.
+
+use super::*;
+use crate::engines::Engine;
+use crate::metrics::{directory_size_bytes, process_rss_bytes, LatencyStats, Throughput};
+use serde_json::json;
+use std::time::Instant;
+
+const COMPONENT: &str = "blockchain_node";
+const VALUE_BYTES: usize = 64; // representative trie node value
+
+/// Run every state-trie scenario for one engine.
+pub fn run_all(engine: Engine, context: &Context) -> Vec<Record> {
+    vec![
+        block_import(engine, context),
+        random_read(engine, context),
+        pruning(engine, context),
+        sustained(engine, context),
+    ]
+}
+
+/// Insert `count` random keys, returning the keys in insertion order.
+fn populate_keys(
+    store: &mut Box<dyn crate::engines::KvStore>,
+    count: usize,
+    rng: &mut ChaCha8Rng,
+) -> Vec<Vec<u8>> {
+    let mut keys = Vec::with_capacity(count);
+    let mut batch = Vec::with_capacity(1024);
+    for _ in 0..count {
+        let key = hash_key(rng);
+        let value = value_of(rng, VALUE_BYTES);
+        keys.push(key.clone());
+        batch.push((key, value));
+        if batch.len() == 1024 {
+            store.commit_batch(&batch, false);
+            batch.clear();
+        }
+    }
+    if !batch.is_empty() {
+        store.commit_batch(&batch, false);
+    }
+    store.flush();
+    keys
+}
+
+/// Block import: durable batched commits, one batch per block.
+fn block_import(engine: Engine, context: &Context) -> Record {
+    let blocks = context.scaled(1_000);
+    let writes_per_block = 200; // state changes per block
+    let directory = context.fresh_directory("trie_block_import");
+    let mut store = engine.open(&directory);
+    let mut rng = ChaCha8Rng::seed_from_u64(context.seed);
+
+    let mut block_durations = Vec::with_capacity(blocks);
+    let mut bytes = 0u64;
+    let mut ops = 0u64;
+    let start = Instant::now();
+    for _ in 0..blocks {
+        let mut batch = Vec::with_capacity(writes_per_block);
+        for _ in 0..writes_per_block {
+            let key = hash_key(&mut rng);
+            let value = value_of(&mut rng, VALUE_BYTES);
+            bytes += (key.len() + value.len()) as u64;
+            batch.push((key, value));
+            ops += 1;
+        }
+        let started = Instant::now();
+        store.commit_batch(&batch, true); // durable block commit
+        block_durations.push(started.elapsed());
+    }
+    let elapsed = start.elapsed();
+    store.flush();
+    let disk_bytes = directory_size_bytes(&directory);
+    drop(store);
+
+    let mut record = Record::new(COMPONENT, "block_import", engine.name());
+    record.params = json!({ "blocks": blocks, "writes_per_block": writes_per_block, "value_bytes": VALUE_BYTES, "sync": true });
+    record.throughput = Some(Throughput::new(ops, bytes, elapsed));
+    record.latency = Some(LatencyStats::from_durations(block_durations)); // per-block commit
+    record.disk_bytes = Some(disk_bytes);
+    record
+}
+
+/// Random point reads over a large state, cold (after reopen) and warm.
+fn random_read(engine: Engine, context: &Context) -> Record {
+    let count = context.scaled(1_000_000);
+    let reads = context.scaled(100_000);
+    let directory = context.fresh_directory("trie_random_read");
+
+    let keys = {
+        let mut store = engine.open(&directory);
+        let mut rng = ChaCha8Rng::seed_from_u64(context.seed);
+        let keys = populate_keys(&mut store, count, &mut rng);
+        drop(store);
+        keys
+    };
+
+    // Reopen for a cold cache.
+    let store = engine.open(&directory);
+    let mut rng = ChaCha8Rng::seed_from_u64(context.seed ^ 0x5151);
+    let pick_key =
+        |rng: &mut ChaCha8Rng| -> &Vec<u8> { &keys[(rng.next_u64() as usize) % keys.len()] };
+
+    let cold_count = reads.min(10_000);
+    let mut cold = Vec::with_capacity(cold_count);
+    for _ in 0..cold_count {
+        let key = pick_key(&mut rng);
+        let started = Instant::now();
+        let got = store.get(key);
+        cold.push(started.elapsed());
+        assert!(got.is_some(), "missing key during random_read");
+    }
+    let mut warm = Vec::with_capacity(reads);
+    for _ in 0..reads {
+        let key = pick_key(&mut rng);
+        let started = Instant::now();
+        let _ = store.get(key);
+        warm.push(started.elapsed());
+    }
+    drop(store);
+
+    let cold_statistics = LatencyStats::from_durations(cold);
+    let mut record = Record::new(COMPONENT, "random_read", engine.name());
+    record.params = json!({ "keys": count, "reads": reads });
+    record.latency = Some(LatencyStats::from_durations(warm));
+    record.extra = json!({ "cold_latency": cold_statistics });
+    record
+}
+
+/// Pruning: delete the oldest half of the keys, measure delete latency and the
+/// disk space reclaimed (after a compaction opportunity).
+fn pruning(engine: Engine, context: &Context) -> Record {
+    let count = context.scaled(500_000);
+    let directory = context.fresh_directory("trie_pruning");
+    let mut store = engine.open(&directory);
+    let mut rng = ChaCha8Rng::seed_from_u64(context.seed);
+    let keys = populate_keys(&mut store, count, &mut rng);
+    let disk_before = directory_size_bytes(&directory);
+
+    let prune_count = count / 2;
+    let started = Instant::now();
+    for key in keys.iter().take(prune_count) {
+        store.delete(key);
+    }
+    store.flush();
+    let delete_elapsed = started.elapsed();
+
+    // Give the engine a chance to reclaim (reopen triggers settling; RocksDB
+    // tombstones need compaction, surfaced as residual disk).
+    drop(store);
+    drop(engine.open(&directory));
+    let disk_after = directory_size_bytes(&directory);
+
+    let mut record = Record::new(COMPONENT, "pruning", engine.name());
+    record.params = json!({ "keys": count, "pruned": prune_count });
+    record.throughput = Some(Throughput::new(prune_count as u64, 0, delete_elapsed));
+    record.disk_bytes = Some(disk_after);
+    record.extra = json!({
+        "disk_before_bytes": disk_before,
+        "disk_after_bytes": disk_after,
+        "reclaimed_fraction": 1.0 - (disk_after as f64 / disk_before.max(1) as f64),
+        "delete_total_ms": delete_elapsed.as_secs_f64() * 1e3,
+    });
+    record
+}
+
+/// Sustained interleaved read/write while sampling RSS — surfaces compaction
+/// memory pressure over time.
+fn sustained(engine: Engine, context: &Context) -> Record {
+    let rounds = context.scaled(2_000);
+    let writes_per_round = 100;
+    let reads_per_round = 100;
+    let directory = context.fresh_directory("trie_sustained");
+    let mut store = engine.open(&directory);
+    let mut rng = ChaCha8Rng::seed_from_u64(context.seed);
+
+    let rss_before = process_rss_bytes();
+    let mut max_rss = rss_before.unwrap_or(0);
+    let mut keys: Vec<Vec<u8>> = Vec::new();
+    let mut ops = 0u64;
+    let start = Instant::now();
+    for round in 0..rounds {
+        let mut batch = Vec::with_capacity(writes_per_round);
+        for _ in 0..writes_per_round {
+            let key = hash_key(&mut rng);
+            keys.push(key.clone());
+            batch.push((key, value_of(&mut rng, VALUE_BYTES)));
+            ops += 1;
+        }
+        store.commit_batch(&batch, false);
+        if !keys.is_empty() {
+            for _ in 0..reads_per_round {
+                let key = &keys[(rng.next_u64() as usize) % keys.len()];
+                let _ = store.get(key);
+                ops += 1;
+            }
+        }
+        if round % 100 == 0 {
+            if let Some(rss) = process_rss_bytes() {
+                max_rss = max_rss.max(rss);
+            }
+        }
+    }
+    let elapsed = start.elapsed();
+    store.flush();
+    drop(store);
+
+    let mut record = Record::new(COMPONENT, "sustained", engine.name());
+    record.params = json!({ "rounds": rounds, "writes_per_round": writes_per_round, "reads_per_round": reads_per_round });
+    record.throughput = Some(Throughput::new(ops, 0, elapsed));
+    record.rss_delta_bytes = rss_before.map(|before| max_rss.saturating_sub(before));
+    record.extra = json!({ "peak_rss_bytes": max_rss });
+    record
+}
--- a/benchmarks/db-bench/src/workloads/storage.rs
+++ b/benchmarks/db-bench/src/workloads/storage.rs
@@ -1,0 +1,267 @@
+//! Storage Provider workloads — model the post-Issue-#100 per-bucket database:
+//! each bucket is one small, independent store keyed by MMR leaf position.
+
+use super::*;
+use crate::engines::Engine;
+use crate::metrics::{
+    directory_size_bytes, open_fd_count, process_rss_bytes, LatencyStats, Throughput,
+};
+use serde_json::json;
+use std::time::Instant;
+
+const COMPONENT: &str = "storage_provider";
+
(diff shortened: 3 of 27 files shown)


benchmarks/db-bench/src/workloads/storage.rs
```diff
@@ -0,0 +1,267 @@
+//! Storage Provider workloads — model the post-Issue-#100 per-bucket database:
+//! each bucket is one small, independent store keyed by MMR leaf position.
```
mudigal, line 2: Nice modeling. One thing to fold in later: the real store keys nodes/chunks by random H256 content hash (storage/mod.rs), not sequential position — so keying these scenarios by hash would better match real access locality.

benchmarks/db-bench/src/workloads/storage.rs
```diff
@@ -0,0 +1,267 @@
...
+    while (position as usize) < total {
+        let mut batch = Vec::with_capacity(batch_size);
+        for _ in 0..batch_size {
+            if (position as usize) >= total {
+                break;
+            }
+            let value = value_of(&mut rng, value_size);
+            bytes += (value.len() + 8) as u64;
+            batch.push((position_key(position), value));
+            position += 1;
```
mudigal, line 97: Append-only here — would be great to add an overwrite/update variant eventually, since the mutable/CDC design churns and orphans nodes, and that's the write+GC behavior that most differentiates the engines.

benchmarks/db-bench/src/workloads/storage.rs
```diff
@@ -0,0 +1,267 @@
...
+    // Reopen for a cold-ish cache, then measure warm random reads.
+    let store = engine.open(&directory);
+    let mut rng = ChaCha8Rng::seed_from_u64(context.seed ^ 0xABCD);
+
+    // Cold phase (first pass).
+    let mut cold = Vec::with_capacity(reads.min(5_000));
+    for _ in 0..reads.min(5_000) {
+        let position = rng.next_u64() % leaf_count as u64;
+        let started = Instant::now();
+        let got = store.get(&position_key(position));
```
mudigal, line 131: Worth noting `proof_read` is a single point lookup; a real MMR proof reads `O(log n)` sibling/peak nodes, so this is a floor on proof cost. A tree-walk variant would make it representative.

benchmarks/db-bench/src/workloads/state_trie.rs
```diff
@@ -0,0 +1,222 @@
...
+    let mut rng = ChaCha8Rng::seed_from_u64(context.seed);
+
+    let mut block_durations = Vec::with_capacity(blocks);
+    let mut bytes = 0u64;
+    let mut ops = 0u64;
+    let start = Instant::now();
+    for _ in 0..blocks {
+        let mut batch = Vec::with_capacity(writes_per_block);
+        for _ in 0..writes_per_block {
+            let key = hash_key(&mut rng);
```
mudigal: Fresh key per write, so no in-place updates — an overwrite scenario would exercise the compaction/rewrite paths where RocksDB and ParityDB differ most. Good candidate for a follow-up.

docs/design/database-evaluation/03-configuration-guide.md
```diff
@@ -0,0 +1,156 @@
...
+layout.)
+
+> Note: the original plan attributed this to Issue #65, but Issue #65 is the
+> "Robust Syncing Protocol for Dynamic Primary and Replica Node Topologies" — a
+> different topic. Track key-prefix restructuring under Issue #101 (or a new
+> dedicated issue), not under #65.
+
+---
+
+## OS-level checklist
```
mudigal: The pool cap limits how much open buckets cost. But nothing limits how many buckets exist in total, and total buckets have their own costs even when closed:

1. Disk - an empty bucket still takes ~72 KiB on disk. A million buckets = ~69 GiB before storing any real data. Stake covers bytes stored, not the number of buckets, so tiny buckets add this cost for free.
2. Files - every bucket is 2 files. At a million buckets the buckets/ folder gets too big and needs subfolders.
3. Restart time - after a crash the node checks every bucket on startup. More buckets = slower recovery, while the challenge clock is running.

Two suggestions: add a "**max number of buckets**" line to the checklist (worked out from disk, file, and restart budgets), and require a minimum agreement size so even an empty bucket pays for itself. Not a blocker - just a gap in the sizing story.

mudigal (changes requested):
Thanks @franciscoaguirre — this has come a long way. From our chat:

1. Cleanup. Rebase onto dev; re-point the docs' code references (#322 moved storage to crates/providers/storage/ — your findings still hold there). Move the write-up to docs/research/database-evaluation/; docs/design/ is for ratified specs. Keep the harness checked in.

2. **SQLite yes - one DB per bucket, not two**. Start with the single-file variant you called out in doc 05: one WAL gives the write-ordering guarantee for free, no choke-point function, half the FDs. We knowingly accept the `page_size` compromise; the split stays available later - as you showed, it's bucket-local and reversible.

3. Spec the in-bucket prune path. Live buckets shrink too (delete_before, version churn), and chunks are shared within a bucket - the design needs to say when a chunk is safe to remove and how `used_bytes` comes down. Doc 03's vacuum checklist already assumes this exists.

4. Add provider capacity bounds. Max N buckets per provider, min M bytes per bucket - gives the pool/FD/floor math a hard denominator and rules out millions of near-empty buckets. Stake bounds bytes; this bounds count.

5. File the bugs as issues - the `used_bytes` under-charge first (live quota bypass today) - and link them from the doc.

CLAUDE.md
```diff
@@ -643,6 +643,7 @@ For the full review criteria (Parity Standards), see the `/review` skill. The re
 | [Storage Marketplace](docs/design/marketplace.md) | Provider capacity & discovery |
 | [Checkpoint Protocol](docs/design/CHECKPOINT_PROTOCOL.md) | Automated checkpoint management |
 | [File System Architecture](docs/filesystems/ARCHITECTURE.md) | Layer 1 encoding, security, blockchain details |
+| [Database Engine Evaluation](docs/design/database-evaluation/README.md) | Benchmark-backed DB selection |
```
bkontur: all this database-evaluations should go to the `docs/drafts`, `docs/design/*` is for approved canonicalized design stuff

danielbui12:
> Thanks @franciscoaguirre — this has come a long way. From our chat:
> 
> 1. Cleanup. Rebase onto dev; re-point the docs' code references (#322 moved storage to crates/providers/storage/ — your findings still hold there). Move the write-up to docs/research/database-evaluation/; docs/design/ is for ratified specs. Keep the harness checked in.
> 2. **SQLite yes - one DB per bucket, not two**. Start with the single-file variant you called out in doc 05: one WAL gives the write-ordering guarantee for free, no choke-point function, half the FDs. We knowingly accept the `page_size` compromise; the split stays available later - as you showed, it's bucket-local and reversible.
> 3. Spec the in-bucket prune path. Live buckets shrink too (delete_before, version churn), and chunks are shared within a bucket - the design needs to say when a chunk is safe to remove and how `used_bytes` comes down. Doc 03's vacuum checklist already assumes this exists.
> 4. Add provider capacity bounds. Max N buckets per provider, min M bytes per bucket - gives the pool/FD/floor math a hard denominator and rules out millions of near-empty buckets. Stake bounds bytes; this bounds count.
> 5. File the bugs as issues - the `used_bytes` under-charge first (live quota bypass today) - and link them from the doc.

@mudigal @franciscoaguirre quick catch up. so we'll use SQLite one database per bucket, and another global database to store buckets' metadata, provider stats, ... is and SQLite as well?

bkontur:
@franciscoaguirre @mudigal Could we have some "smoldocs" or "result" or "decissions" or "open questions" or "actual state" section in the PR desciption? 

This is pretty heavy topic underneath, I am just thinking if we should split this to two topics/designs:
- one database per bucket
- change rocksdb to SQLite db (would this still make sense without changing to "one db per bucket"?)

bkontur:
> @mudigal @franciscoaguirre quick catch up. so we'll use SQLite one database per bucket, and another global database to store buckets' metadata, provider stats, ... is and SQLite as well?

@danielbui12 good catch, this is exactly also one of my questions, which I haven't pushed yet:

We have couple of cross-bucket calls/RPCs, e.g.: 

- GET /buckets → list_buckets().
- GET /stats → get_bucket_stats(), total_bytes(), total_nodes(): per-bucket used_bytes plus a count of all chunks.

_(Probably, we cannot remove those RPCs)_

So basically, for most of the write operations, we will need to access two DBs (bucket one and metadata/summary one), which brings some questions/issues:
- concurrent writes need to access the shared metadata/summary DB, what problem does this bring? Do we need to reintroduces the shared-DB single-writer bottleneck or some async solution?
 - how to ensure consistency between writes (bucket-one and summary) in case something crashes between this two writes (recalculate on startup or something?)

bkontur:
One more question, looks like not covered by the benchmarks: **hot bucket**. Imagine one .dot site (SPA) that, lots of people load at the same time while the owner (one writer) uploads a new version: **100, 1k, 10k concurrent downloads** while **uploading** to the **same bucket**.

How would this behave (locks, connections, latencies)? Any bottlenecks? Would it scale?

Something like adding more measurements on **one bucket file**:
   - N concurrent readers (N = 100, 1k, 10k), read p50/p99
   - N readers + 1 writer for a sustained period
   - 100+ contending writers (separate topic https://github.com/paritytech/web3-storage/pull/344)
   - read p99 during a large checkpoint

bkontur (changes requested):
Very nice and solid work here - the benchmarking and comparisions.

I would just make the design clear and sharp about "one DB per bucket" vs. the DB engine. The first is the architecture we build the provider around and should outlive whatever engine we pick. SQLite fits it well today, no objection, but it should stay a replaceable detail under a stable per-bucket contract. I see these as a two separate orthogonal things, that should be stated clearly (I would also implement this as a two separate orthogonal PRs). What if SQLite-or-whatever-db we choose goes unmaintained and we would need to switch to other DB?

Also the multi-bucket reads (https://github.com/paritytech/web3-storage/pull/253#issuecomment-5599987081) are a good example of why: /buckets and /stats are an architecture question (per-bucket files plus a provider-level summary, and how they stay consistent), not an engine question.

A [short](https://github.com/paritytech/web3-storage/pull/253#issuecomment-5599599725) "decisions / open questions / current state" section in the PR description would help, the thread is getting long :)

mudigal:
> > @mudigal @franciscoaguirre quick catch up. so we'll use SQLite one database per bucket, and another global database to store buckets' metadata, provider stats, ... is and SQLite as well?
> 
> @danielbui12 good catch, this is exactly also one of my questions, which I haven't pushed yet:
> 
> We have couple of cross-bucket calls/RPCs, e.g.:
> 
> * GET /buckets → list_buckets().
> * GET /stats → get_bucket_stats(), total_bytes(), total_nodes(): per-bucket used_bytes plus a count of all chunks.
> 
> _(Probably, we cannot remove those RPCs)_
> 
> So basically, for most of the write operations, we will need to access two DBs (bucket one and metadata/summary one), which brings some questions/issues:
> 
> * concurrent writes need to access the shared metadata/summary DB, what problem does this bring? Do we need to reintroduces the shared-DB single-writer bottleneck or some async solution?
> * how to ensure consistency between writes (bucket-one and summary) in case something crashes between this two writes (recalculate on startup or something?)

@bkontur @danielbui12 

tl;dr: Nothing authoritative needs a shared database, so neither problem arises.

My thoughts on the 3 points below - 

1. On the two RPCs: they're already going away - #396 removes `GET /buckets` outright (it had no callers, and per-bucket enumeration would sidestep the new Layer-0 read gates) and cuts `GET /stats` to aggregate totals (the dashboard only reads `total_bytes`). The internal `list_buckets()` consumer shrinks to `bucket_ids()`. So the whole cross-bucket surface after that PR is: `bucket_ids()` plus two counters.

2. In the sharded model: `bucket_ids()` is a directory listing - one file per bucket means the filesystem is the index. The authoritative `used_bytes` lives in each bucket's own meta row, written in the same transaction as the bucket's data. The global totals are derived observability numbers: keep them as in-memory counters (or an advisory row in provider.sqlite, which exists anyway for the nonce high-water mark) and recompute them during the startup scrub doc 05 already requires. So writes touch exactly one database, there's no shared-writer contention, and crash consistency for the summary is "recalculate on startup" - fine precisely because the summary is never load-bearing.

3. On bucket metadata generally: the bucket↔drive and bucket↔S3 mappings aren't provider-local at all - they're on-chain (drive-registry's BucketToDrive, s3-registry's S3BucketInfo.layer0_bucket_id). The Layer-1 indexes (fs_indices/, s3_indices/) are per-bucket derived files, so they shard cleanly too. The deeper question your comment touches - which metadata is authoritative vs derived - turned out to deserve its own design: #410 proposes making the key→content mapping committed data in the bucket itself, which formally demotes the provider's index to a rebuildable cache. Please take a look at this.

Tried to keep it short, but got long as was trying to address many points together.

@eskimor  - FYI

benchmarks/db-bench/src/workloads/storage.rs
```diff
@@ -0,0 +1,658 @@
...
+    let count = context.scaled(50_000);
+
+    // Strategy A: delete every key inside the engine.
+    let directory_a = context.fresh_directory("provider_delete_keys");
+    populate(engine, &directory_a, count, 48, context.seed);
+    let mut store = engine.open(&directory_a);
+    let started = Instant::now();
+    for position in 0..count as u64 {
+        store.delete(&position_key(position));
+    }
```
ilchu: I'd say it would be more realistic to batch the deletes, as it is usually done in a prune behind the scenes. Otherwise the actual overhead is with fsyncs and commits on every key deletion.

docs/design/database-evaluation/01-storage-provider-benchmark.md
```diff
@@ -0,0 +1,769 @@
...
+| libmdbx | 100 | 500 | **1000** | **1.00** |
+
+Two results, one of which corrects an assumption this evaluation had been
+carrying:
+
+- **libmdbx spawns one transaction-manager thread per environment**, exactly 1:1.
+  At a thousand open buckets that is a thousand threads. It was added *because*
+  it removes LMDB's two liabilities (no `map_size` ceiling, unconditional
+  `MDBX_NOTLS`), and this single property outweighs both. It is the measured
+  counter-example, not a candidate.
```
ilchu: This sounds to me like an artifact of the library. There should be a knob in the options or some other binding that would allow not doing this, as the C library doesn't seem to spawn threads at all. So we could be downplaying the performance of a potentially very fast contestant here.

benchmarks/db-bench/src/engines/mdbx_store.rs
```diff
@@ -0,0 +1,131 @@
...
+    std::fs::create_dir_all(path).expect("mdbx directory");
+    let options = DatabaseOptions {
+        max_tables: Some(1),
+        mode: Mode::ReadWrite(ReadWriteOptions {
+            // Durability is forced explicitly in `flush` so an unsynced batch is
+            // genuinely unsynced (see the module docs).
+            sync_mode: SyncMode::SafeNoSync,
+            // `None` leaves mdbx to choose: no ceiling, no preallocation.
+            min_size: None,
+            max_size: None,
```
ilchu: Quite likely these options in conjunction with the above no sync are what's causing the high reported empty floor and the reported 3x amplification of this DB in the benches.

docs/design/database-evaluation/02-recommendations.md
```diff
@@ -0,0 +1,329 @@
...
+   append-only until it closes. The code that holds a read open is exactly the
+   code that streams a large download or builds a challenge proof.
+2. **No online compaction.** Measured: 3.3 MiB retained after deleting every key,
+   with no reclamation API. Harmless while buckets die by `unlink`; decisive if a
+   bucket is ever cleared in place.
+3. **`MDB_MAP_FULL` is a cliff**, and `map_size × pool_cap` is an address-space
+   budget no other engine imposes — measured at exactly 1 GiB per instance,
+   1001.5 GiB across a 1000-bucket pool. For buckets holding chunked media the
+   per-bucket ceiling has to be large, which makes the pool cap the binding
+   constraint.
```
ilchu: I've found that there's a `mdb_env_set_mapsize` command for LMDB that can be run when no tx is open to readjust.

docs/design/database-evaluation/02-recommendations.md
```diff
@@ -0,0 +1,329 @@
...
+
+**Three LMDB failure modes map directly onto this workload:**
+
+1. **Long-lived read transactions block page reuse.** Per LMDB's own caveats, no
+   space is reclaimed while a read transaction is open, so the file grows
+   append-only until it closes. The code that holds a read open is exactly the
+   code that streams a large download or builds a challenge proof.
+2. **No online compaction.** Measured: 3.3 MiB retained after deleting every key,
+   with no reclamation API. Harmless while buckets die by `unlink`; decisive if a
+   bucket is ever cleared in place.
```
ilchu: SQLite's harness also does the `VACUUM` under an exclusive lock. And the analogue of that for LMDB would be `mdb_copy -c` to a new file. There's still no incremental vacuum compared to the former, sure.

docs/design/database-evaluation/02-recommendations.md
```diff
@@ -0,0 +1,329 @@
...
+Meilisearch) and a small, simple design, which is a real argument, but it is
+"battle-tested" rather than "systematically verified", and upstream is
+effectively in maintenance mode.
+
+**Three LMDB failure modes map directly onto this workload:**
+
+1. **Long-lived read transactions block page reuse.** Per LMDB's own caveats, no
+   space is reclaimed while a read transaction is open, so the file grows
+   append-only until it closes. The code that holds a read open is exactly the
+   code that streams a large download or builds a challenge proof.
```
ilchu: Same is true for SQLite's WAL, and the mitigation is to just read in chunks and close txs more often, so this is not quite a factor.

docs/design/database-evaluation/03-configuration-guide.md
```diff
@@ -0,0 +1,176 @@
...
+  `-512` (512 KiB) if you expect many simultaneously-hot buckets. The commitment
+  working set is small enough to stay cached; chunk reads are inherently cold,
+  no cache policy helps them, and with `mmap_size` on they are served from the
+  mapping rather than through the pager cache.
+- **Eviction = close.** Closing a connection releases its WAL/shm FDs and cache;
+  reopen is 87 µs on disk, so aggressive eviction is cheap.
+- **Checkpoint on eviction.** Run `PRAGMA wal_checkpoint(TRUNCATE)` before closing
+  a bucket to keep the `-wal` file from growing unbounded across sessions. Chunk
+  ingest is what grows the WAL, so `wal_autocheckpoint` fires on content volume;
+  long-running downloads (read transactions) delay WAL reset for the whole file.
```
ilchu: Here it mentions the same drawback as [above](https://github.com/paritytech/web3-storage/pull/253/changes#r4013799000).

docs/design/database-evaluation/01-storage-provider-benchmark.md
```diff
@@ -0,0 +1,769 @@
...
+The split index was expected to *cost* something on the hit path — one extra hop
+from index to row. It does the opposite: the index B-tree is now small enough to
+stay cached, so hits get 4.8× faster too, at identical space.
+
+**What this does to the engine comparison.** Against pass 5's untuned SQLite:
+
+| Metric | SQLite as measured | SQLite tuned (mmap + split) | LMDB | Remaining gap |
+|---|---:|---:|---:|---:|
+| Dedup miss p50 | 746 µs | **3.92 µs** | 0.50 µs | 7.8× |
+| Chunk read warm p50 | 711 µs | **34.58 µs** | 19 µs | 1.8× |
```
ilchu: AFAICT the LMDB numbers here come from the `final-run.json`, whereas the new SQLite numbers are `dedup-experiment.json`. So it's not quite apples to apples, and even somewhere above it was noted that the variation with the same env options between different runs can be more than 1.5x. I think it would make sense to rerun the LMDB for comparison too.

benchmarks/db-bench/src/engines/heed_store.rs
```diff
@@ -0,0 +1,148 @@
...
+            // MDB_NOTLS: read transactions must not consume a thread-local
+            // storage key, or many open environments exhaust the process's
+            // supply and LMDB starts returning MDB_TLS_FULL.
+            .read_txn_without_tls()
+            .map_size(map_size())
+            .max_dbs(1)
+            .max_readers(128)
+            // Durability is forced explicitly in `flush`, so that an unsynced
+            // batch is genuinely unsynced (see the module docs).
+            .flags(EnvFlags::NO_SYNC | EnvFlags::NO_META_SYNC)
```
ilchu: I think it doesn't make sense to force `NO_SYNC` for LMDB specifically. Whereas we try to turn off this option for any other DB being tested where applicable, in most cases those configurations are still legitimately production usable. I.e. they guarantee that some recent transactions may be dropped, but the file remains readable and consistent. On the other hand, LMDB explicitly advises against turning sync off as it may corrupt the file unless the file system orders reads properly, which can't be guaranteed. We even quote it [here](https://github.com/paritytech/web3-storage/pull/253/changes#diff-e88f8ffe84a2c7087cd42515c00a1cbe58194bc5becb4d648c888f4e783ed650R244). So maybe we should actually run it with only meta sync off?

docs/design/database-evaluation/03-configuration-guide.md
```diff
@@ -0,0 +1,176 @@
...
+
+```sql
+-- One file per bucket: {storage_path}/buckets/{bucket_id}.sqlite
+PRAGMA page_size    = 32768;      -- set FIRST, before WAL; see below
+PRAGMA journal_mode = WAL;
+PRAGMA synchronous  = NORMAL;     -- the default for content transactions; raised
+                                  -- to FULL around each commitment transaction
+PRAGMA busy_timeout = 5000;
+PRAGMA cache_size   = -2000;      -- ~2 MiB; the commitment working set is ~10 MB per 100 k uploads
+PRAGMA mmap_size    = 1073741824; -- 1 GiB: worth 4.3x on chunk reads, see below
```
ilchu: These pragmas prescribe both custom page and mmap size, but the actual runs did either the mmap only, default 4 KiB pages (in `dedup-experiment.json`) or 32 KiB pages, no mmap (in `sqlite-page-size-run.json`). I'd be hesitant to assume that the stated 1.58x and 4.8x gains compose, especially given overall fiddliness of huge pages.

ilchu (changes requested):
It's an enormous amount of work - kudos for making it happen. I still have to admit that it's really hard to grasp or understand the details without the assistant, so I tried to do my best 😅 

I do feel that there's sufficient backing for the shared vs sharded conclusion, but less so for the exact choice of the DB. I don't have anything against SQLite per se, and my gut says it's probably the most reasonable solution. Still, both LMDB and MDBX have been downplayed not by their intrinsic properties, but by some artifacts of setup, and the comparison has not always completely been apples to apples. Details below.

So ideally I'd see some run of mdbx vs LMDB vs SQLite in this PR as the last definitive test. The first DB without the 1:1 threadmapping and with durable sync to avoid amplification, the second with only `NO_META_SYNC` on, and the third with bigger pages + mmap size + all the other recommended pragmas all on at once, and then analyze those final numbers only.

